PR #42, opened 2014-04-03
Dynamic commanded states

taldcroft:
This adds support for dynamically constructed commanded states in Kadi.  The current branch is not complete and should not be merged.

To do:
- [x] Narrative docs
- [x] ~~Frequent command updates and transfer to GRETA network~~.  To be done in #107.
- [x] Command line script
- [x] Consider improving / simplifying callback machinery (make transition be `key` = class name and `val` = function (using a partial enclosure if needed for args, but not likely since function gets command already). 
- [x] Document base transition classes
- [x] Transitions for Larry
- [x] Reduce states per cmd_states
- [x] Look at different backstop command data structures and normalize.  Replace CmdList with BackstopTable?
- [x] ~~Don't break state if there are no `trans_keys`.  See TODO in code.  The sun vector transitions that occur during a maneuver generate a null transition.  Changing `if date != datestarts[-1]:` to `if date != datestarts[-1] and state.trans_keys:` broke all the tests (not clear why).~~ This is a bit of a pain and probably not worth the effort since default is to reduce these states away.
- [x] ~~Possibly rename `type` to `class`?  (ORBPOINT has `TYPE` param that conflicts).~~  Probably overkill.
- [x] Add some class method hook to allow for custom initialization of transitions.  Initial use case is to bundle `add_sun_vector_transitions` into a transition class `SunVectorTransition`.
- [x] Remove unnecessary `type` filters when `tlmsid` is enough

Closes https://github.com/sot/cmd_states/issues/25


jeanconn:
In general, are we still going to be fetching cmds from backstop files using the timelines database as a reference?  If so, should we finally make a timelines module with a defined api to get maps of times to files instead of every project accessing whichever database directly?  Seems like that would also help with dynamic timelines and testing.   It also might be a win to make validate_states into a useful module.

taldcroft:
> In general, are we still going to be fetching cmds from backstop files using the timelines database as a reference? 

That is not the idiom I have in mind.  Users would be getting commands strictly from the kadi cmds database, which already has the timelines filtering applied.  The kadi cmds database will be frequently updated to reflect changes to timelines.

E.g. most of the following [getting timelines manually and using this to get commands](https://github.com/acisops/acis_thermal_check/blob/1d60317/acis_thermal_check/state_builder.py#L139) should go away.  Understood that this is a bit of a leap for a core flight tool, but that would be the plan.

Of course `kadi/update_cmds.py` will be doing what you just said, but that's fine since that is an "internal" application.

PR #66, opened 2015-03-06
Startup issues

taldcroft:
This fixes a couple of annoyances/problems when importing kadi.


taldcroft:
@jeanconn

taldcroft:
I don't remember why this didn't merged.

taldcroft:
Rebased and tested that this works if there is no secret key file and the directory where it would live is not writeable.

I think this didn't get merged because of a workaround to make the production version of the secret key world readable.  We could change that when this is merged.

PR #68, opened 2015-05-29
Fix task_schedule user and add an OCC version

taldcroft:
Note that having the `--ftp` flag is OK even though the currently installed OCC version does not.


jeanconn:
Not sure where to note that a post-install step for the install of this in /proj/sot/ska/test on GRETA will be to update the crontab to call the new occ version.

PR #70, opened 2015-07-15
WIP: Add load_seg_changes app to kadi

jeanconn:
Add app to look at htmldiff of net changes to iFOT load segments file during last week.

### DIFF
--- a/kadi/settings.py
+++ b/kadi/settings.py
@@ -64,7 +64,7 @@ INSTALLED_APPS = (
     'kadi.events',
     'mica.web',
     'find_attitude.web',
-    'load_seg_changes.web',
+    'timelines.load_seg_changes.web',
 )
 
 MIDDLEWARE_CLASSES = (
--- a/kadi/settings.py
+++ b/kadi/settings.py
@@ -64,6 +64,7 @@ INSTALLED_APPS = (
     'kadi.events',
     'mica.web',
     'find_attitude.web',
+    'load_seg_changes.web',
 )
 
 MIDDLEWARE_CLASSES = (
--- a/kadi/templates/base.html
+++ b/kadi/templates/base.html
@@ -42,6 +42,7 @@
           <ul class="dropdown">
             <li> <a href="/pcad_acq">PCAD Acquisition Table</a> </li>
             <li> <a href="/find_attitude">Find attitude in BSH</a> </li>
+            <li> <a href="/load_seg_changes">This Week's Load Changes</a></li>
           </ul>
         </li>
       </ul>
--- a/kadi/urls.py
+++ b/kadi/urls.py
@@ -4,6 +4,7 @@ from django.contrib import admin
 import kadi.views
 import mica.web.views
 import find_attitude.web.views
+import load_seg_changes.web.views
 
 admin.autodiscover()
 
@@ -15,6 +16,7 @@ urlpatterns = patterns('',
                        url(r'^mica/$', mica.web.views.IndexView.as_view()),
                        url(r'^pcad_acq/$', mica.web.views.AcqView.as_view()),
                        url(r'^find_attitude/$', find_attitude.web.views.index),
+                       url(r'^load_seg_changes/$', load_seg_changes.web.views.LoadSegView.as_view()),
                        )
 
 # Another way to do this, corresponds to commented code in kadi/views.py
--- a/kadi/urls.py
+++ b/kadi/urls.py
@@ -4,7 +4,7 @@ from django.contrib import admin
 import kadi.views
 import mica.web.views
 import find_attitude.web.views
-import load_seg_changes.web.views
+import timelines.load_seg_changes.web.views
 
 admin.autodiscover()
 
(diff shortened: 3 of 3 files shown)


taldcroft:
:+1: on the idea.  

Do you think the underlying functionality could be provided within an existing package like timelines?  I realize that is not actually an installed package, but of course it could/should be.

taldcroft:
Relatedly, there are several times recently when I just wanted to see the as-run and as-scheduled list of obsids and targets.  Sort-of what Replan Central provides, but with more like -7 to +21 days of context.

For mega-bonus points, show a diff of that view based on how it looked on date X and date Y (which Y defaulting to NOW).  We just need a schedule_time_machine that keeps a git repo of the schedule each hour or so.

jeanconn:
On Wed, Jul 15, 2015 at 2:29 PM, Tom Aldcroft <notifications@github.com>
wrote:

> [image: :+1:] on the idea.
>
> Do you think the underlying functionality could be provided within an
> existing package like timelines? I realize that is not actually an
> installed package, but of course it could/should be.
>

We don't actually have the data products in the timelines database for
this, so it doesn't really help us as a package/product.  In timelines we
have the current record of what we think happened, but we aren't saving
what the snapshot of things was last week, for example.

The files I want are just in the history in the time machine that we
currently keep in arc, so this seemed distinct and lightweight enough to
deserve its own package.

As for why I did this, every time I end up with something that doesn't make
sense in the timeline time, I end up looking at the arc files myself and
comparing them with iFOT and trying to figure out what they ingested and
why it broke things.  I just scripted up that process for this.  Did it
again yesterday as a first step when I didn't understand the cmd states
validation (before I realize that the table had no new data at all).

jeanconn:
> Relatedly, there are several times recently when I just wanted to see the
> as-run and as-scheduled list of obsids and targets. Sort-of what Replan
> Central provides, but with more like -7 to +21 days of context.
>
> For mega-bonus points, show a diff of that view based on how it looked on
> date X and date Y (which Y defaulting to NOW). We just need a
> schedule_time_machine that keeps a git repo of the schedule each hour or so.
>

Yes, that's been on my mind as well.  I'd also like the plots from the
validation tools to be interactive and let me click through to mica for
obsid info.

taldcroft:
Even though the current `timelines` doesn't own all the data products, your new functionality (and my secondary feature request) are precisely about the Chandra timelines, so it seems like a pretty good match.  The schedule time machine data could sensibly live in `$ska/data/timelines`.

I can't exactly justify it, but I'm not a huge fan of micro-packages that have no room to grow (because of the name).   But `timelines.load_seg_changes` would be perfect.  I guess I like hierarchical organization.

jeanconn:
Oh.  I thought we wanted more micro-packages so we could just install them and get stuff done without having to cut and approve new releases of already in-use packages.

jeanconn:
Also, wouldn't the obsid list and such you requested be a feature request on cmd_states not timelines?

taldcroft:
It would need to be installed through skare via the usual release mechanism, and approving a new module within an existing package is no additional effort, so the "getting stuff done" isn't necessarily faster with micro-packages.

Since they run from the same cron job I make little distinction between timelines and cmd_states.  I picked `timelines` because that name has such a tight connection to the concept, namely providing information about alternate mission timelines.

taldcroft:
Added a commit to get the working via timelines in a local test area.

taldcroft:
Having seen this run, I'm trying to understand in detail what I'm seeing.  I guess what I was expecting was two columns:

- Left : the approved command loads in the database at the start date.
- Right : the as-run command loads as of the stop date, starting at the start date.

Putting in Start = 2015:180:16:21:12.533 and Stop = 2015:197:16:21:12.533 gives results that I find difficult to interpret.

jeanconn:
I don't actually know what you mean for  " the as-run command loads as of
the stop date, starting at the start date."

My intent was to see the diffs in just the iFOT file between time1 and
time2 without any filtering of that file.  In the rdb file, loads are added
and loads are removed off the top as they run out of scope for your script
in arc.  You could filter to a common start time, but then you'd be
throwing away part of the pure diff that might actually help you notice a
problem.

Using database snapshots from specified times covering specified time
ranges would be a very different tool.

taldcroft:
> I  don't actually know what you mean for  " the as-run command loads as of
the stop date, starting at the start date."

I mean, e.g. on Monday the 7th there are a certain set of approved command loads that are either already uplinked or scheduled for uplink over the coming week.  OK, next Monday the 14th rolls around and there was another V404 TOO on Thursday so from Friday to Monday 14th we ran some different loads.  So in my mind this tool would provide the diffs from the expected schedule from 7th to 14th vs. the actual schedule that happened over that same period (which is now in the past so we know).

jeanconn:
​I don't really understand your example or what you'd need for data to
answer that question.  What's your reference point for the expected
schedule?

The problem I've been trying to answer is the answer for quick inspection
of when the "now in the past so we know" is obviously incorrect because
something wasn't ingested or overwrote something else.  However, using this
to look at some more complicated diffs, another problem is that diff is
being a little too generous about matches when segments are removed
observing-only.  So maybe to even help me with my simple problems we need
what you want.

jeanconn:
Moved to [https://github.com/sot/ska-projects/issues/33](https://github.com/sot/ska-projects/issues/33)

PR #81, opened 2016-05-09
Add mica star history app

jeanconn:
Add mica star history app

### DIFF
--- a/kadi/templates/base.html
+++ b/kadi/templates/base.html
@@ -41,6 +41,7 @@
         <li class="has-dropdown"> <a href="/"> Utilities</a>
           <ul class="dropdown">
             <li> <a href="/pcad_acq">PCAD Acquisition Table</a> </li>
+            <li> <a href="/star_hist">Star History</a> </li>
             <li> <a href="/find_attitude">Find attitude in BSH</a> </li>
           </ul>
         </li>
--- a/kadi/urls.py
+++ b/kadi/urls.py
@@ -14,6 +14,7 @@ urlpatterns = patterns('',
                        url(r'^$', kadi.views.IndexView.as_view()),
                        url(r'^mica/$', mica.web.views.IndexView.as_view()),
                        url(r'^pcad_acq/$', mica.web.views.AcqView.as_view()),
+                       url(r'^star_hist/$', mica.web.views.StarHistView.as_view()),
                        url(r'^find_attitude/$', find_attitude.web.views.index),
                        )
 


taldcroft:
Looks good, at least on paper.

jeanconn:
OK.  Next stages with the "looks good on paper" approval?  Given nothing calls it, can the app just go?

taldcroft:
This needs coordinated testing / review with https://github.com/sot/mica/pull/79.  I'll have a look.

taldcroft:
When the page is loaded initially there is a message `No acq or guide history for star ::`.  That should be suppressed.

taldcroft:
I think an actual `Submit` button would be helpful.  Just hitting return is not 100% intuitive.

taldcroft:
The `obsid` value should be a link to the corresponding mica page.

taldcroft:
`star tracked` needs an explanation.  At this point the `Notes:` should be a bulleted list.

taldcroft:
Is the plan to retire https://icxc.harvard.edu/cgi-bin/aspect/get_stats/get_stats.cgi ?

jeanconn:
Regarding 'Submit', for some reason I thought you didn't like them in the kadi pages (maybe using the mica page as a reference?).  I prefer having them myself and that's not a problem.  Thanks for the feedback!

jeanconn:
And yes, my plan was to begin to plug in this page to replace get_stats.cgi.

taldcroft:
About Submit, I guess I find that if there is just one input box it seems to be clear to hit enter, but with more than one it isn't so much.  My taste can evolve on the time scale of open PRs, but it's not a huge deal.

taldcroft:
`start time` and `stop time` should be capitalized.

taldcroft:
Slightly out of scope, but what about including a table of the corresponding AGASC entry for the star?  This might be relevant/useful in some cases.

taldcroft:
OK, I've finished my review, looks good!

jeanconn:
Regard the table of the AGASC entry, that does make sense.  I suppose I was also thinking about using this tool to replace the "static" pages we have in mica with the star history content:

https://icxc.cfa.harvard.edu/aspect/mica_reports/18/18625/star_397676392.html

and those pages already have the AGASC info, so I should have just brought that over to this display anyway.

taldcroft:
> I suppose I was also thinking about using this tool to replace the "static" pages we have in mica with the star history content

Yes!  And good to see you already have some code and a template to make this table in a nice view.

PR #82, opened 2016-07-19
Change quaternion division to delta quaternion

taldcroft:
Original quaternion division expression gives correct maneuver angle but wrong delta quaternion.  Quaternion 0.4 issues a warning about this incorrect usage.

To do:
- [ ] Update version
- [ ] Test by copying flight events database, dropping some rows, reprocess, confirm same maneuvers.


taldcroft:
In this context the delta quaternion is used only for the maneuver angle.  The existing code gives the correct angle, and the warning will go away with https://github.com/sot/Quaternion/pull/4.

Close when that gets merged.

PR #83, opened 2016-08-29
Update installation notes for mica in production

taldcroft:
The custom /proj/web-kadi install of mica has been moved into HIDE and
is no longer used in production.

### DIFF
--- a/NOTES.web-test-install.rst
+++ b/NOTES.web-test-install.rst
@@ -112,15 +112,7 @@ Restart::
 
 Mica
 ^^^^^
-As needed::
-
-  cd ~/git/mica
-  git branch  # confirm correct web branch
-  git status  # confirm no stray modifications
-
-  ls -ld $PYTHONPATH/mica*
-  rm -rf $PYTHONPATH/mica*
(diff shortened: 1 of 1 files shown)


jeanconn:
Agreed.  Though I wish I could remember why it was off on its own originally.   Was that just because it wasn't in skare?  

I also don't know if you need to restart the server; it seems to pick up new modules and packages in testing without restart.

taldcroft:
> Was that just because it wasn't in skare?

Ding ding ding!  Well, that is quite plausible at least.

> I also don't know if you need to restart the server; it seems to pick up new modules and packages in testing without restart.

Interesting.  In theory the server is supposed to always have a Python spun-up and running, but I have no idea how it works in practice.

jeanconn:
>>  Was that just because it wasn't in skare?
> Ding ding ding! Well, that is quite plausible at least.

It was still installed in /proj/sot/ska even then, but I don't remember the history here.

taldcroft:
Anyway it appears to be working now from flight Ska.

PR #88, opened 2016-12-28
Initial changes for Py3 compatibility

taldcroft:


jeanconn:
:+1:

taldcroft:
@jeanconn - commit 96bb7ec shows an important trick when in converting asciitable to io.ascii.  This showed up in regression testing.  I'm going to include this in an eventual TWiki page on converting code.

PR #94, opened 2017-03-19
Add StarAcq event type and infrastructure for emailed alerts

taldcroft:
This adds a new kadi event that has the number of acq stars ID'd and which ones, along with the size of one-shots.  This is slightly messy and might not end up being merged.  However, it will be run in the short term on the side at 2 hour intervals to provide visibility into metrics related to gyro bias variations and mitigation.  This might belong elsewhere (like attitude_error_mon), but decided the existing infrastructure in kadi was useful for this.

Note also the new flag to use MAUDE instead of CXC as the telemetry data source.

To do:

- [ ] Fix docstring
- [ ] Change email address
- [x] Update task_schedule to only update this model and use MAUDE
- [x] Install and run on the side

@jeanconn - any opinion on the difference between using `localhost` vs. `head.cfa.harvard.edu` for the SMTP server?

### DIFF
--- a/kadi/events/models.py
+++ b/kadi/events/models.py
@@ -22,6 +22,7 @@ utils = None
 
 ZERO_DT = -1e-4
 MAX_GAP = 328.1  # Max gap (seconds) in telemetry for state intervals
+R2A = 206264.8  # radians to arcsec
 
 logger = pyyaks.logger.get_logger(name='events', level=pyyaks.logger.INFO,
                                   format="%(asctime)s %(message)s")
@@ -701,6 +702,8 @@ class TlmEvent(Event):
 
             # Custom processing defined by subclasses to add more attrs to event
             event.update(cls.get_extras(event, event_msidset))
+            if hasattr(cls, 'get_alerts'):
+                event.update(cls.get_alerts(event, event_msidset))
 
             events.append(event)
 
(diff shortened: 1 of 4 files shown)


jeanconn:
I don't really have an opinion on the sending mail choice.  It looks like our original requirement during GAG migration was that local "sendmail" and direct connections to head.cfa.harvard.edu would continue to work.  It looks like direct SMTP connections to localhost are doing the right thing too and we've been moving to that in more and more code.

jeanconn:
For https://github.com/sot/attitude_error_mon/issues/9 I was also planning to add the one shot pitch/yaw breakdown (or a delta quaternion), and I figured it made sense to do it in kadi, so nice to have it done.  Where is this running on the side?

taldcroft:
It's running on /proj/sot/ska/dev-kadi every 2 hours (except not at 2 and 4am, no point).

I'll change alerts to aca_alert, just wanted to make sure it didn't blow up.

kadi/events/models.py
```diff
@@ -1114,6 +1117,112 @@ class Eclipse(TlmEvent):
...
+        Check for large one-shot in pitch or yaw, or a small number of identified acq
+        stars.
+        """
+        def get_obsid():
+            dat = fetch.Msid('cobsrqid', event['tstop'] - 10, event['tstop'] + 300)
+            return int(dat.vals[0])
+
+        alerts = []
+
+        if abs(event['one_shot_pitch']) > 60 or abs(event['one_shot_yaw']) > 60:
```
jeanconn: Is 60 a little low?  I thought the warning on 50220 wasn't that helpful with estimated maneuver error of ~53 arcsec.

```
One shot at 2017:081:20:16:31.008
Roll  : 2.11
Pitch : 63.38
Yaw   : 7.40
```

taldcroft:
I purposely set these limits pretty low for increased awareness in the near term.   But note that 60 is not *that* low from a historical perspective since fewer than 70 of 9000 one-shots in the last 5 years exceed that.  So seeing 60 means something slightly unusual has happened and if we see it every week that means one shots are worse.

The alert on `n_id <= 5` is definitely too jumpy for production.

jeanconn:
I don't immediately see why this kicked off two different notes for the 19504 acquisition:

```
One shot at 2017:121:04:14:26.632
Roll  : 1.65
Pitch : -63.57
Yaw   : 27.50

One shot at 2017:121:04:14:25.898
Roll  : 1.65
Pitch : -63.57
Yaw   : 27.50

```

jeanconn:
I'm seeing events_staracq in /proj/sot/ska/data/kadi/events.db3, so did this end up in flight or was that a previously-existing thing?

taldcroft:
Will leave this as a branch but closing since we have no plans to merge to master.

PR #96, opened 2017-03-29
Add one shot break down (roll, pitch, yaw) to Manvr event

jeanconn:
Add one shot break down (roll, pitch, yaw) to Manvr event

This PR adds the one_shot_roll, one_shot_pitch, and one_shot_yaw to the Manvr class.  A change is also made here to use the AOATTER values in the first guide readout, not the the values that correspond to the time when the RSS of AOATTER2 and AOATTER3 is biggest.

### DIFF
--- a/kadi/events/models.py
+++ b/kadi/events/models.py
@@ -22,6 +22,7 @@ utils = None
 
 ZERO_DT = -1e-4
 MAX_GAP = 328.1  # Max gap (seconds) in telemetry for state intervals
+R2A = 206264.8  # Convert from radians to arcsec
 
 logger = pyyaks.logger.get_logger(name='events', level=pyyaks.logger.INFO,
                                   format="%(asctime)s %(message)s")
@@ -1478,23 +1479,30 @@ class Manvr(TlmEvent):
 
     @classmethod
     @import_ska
-    def get_one_shot(cls, guide_start, one_shot_msid):
+    def get_one_shot(cls, guide_start, aux_msidset):
         """
         Define one_shot attribute (one-shot attitude update following maneuver)
         """
-        # No acq => guide transition so no one-shot defined.  Use -99.0 as a convenience.
+        # No acq => guide transition so no one-shot defined.  Use -99.0 as a convenience for
+        # the one shot length, -999.0 for the axis values.
         if guide_start is None:
-            return -99.0
-
-        t0 = DateTime(guide_start).secs
-        n_times = len(one_shot_msid.times)
-        i0, i1 = np.searchsorted(one_shot_msid.times, [t0 - 10, t0 + 10])
-        if i0 == 0 or i1 == 0 or i0 == n_times or i1 == n_times:
-            raise ValueError('Unable to find complete data for one-shot attitude value'
-                             ' {} {} {} {} {}'
-                             .format(i0, i1, t0, np.min(one_shot_msid.times),
-                                     np.max(one_shot_msid.times)))
-        return np.max(one_shot_msid.vals[i0:i1])
+            return {'one_shot': -99.0,
+                    'one_shot_roll': -999.0, 'one_shot_pitch': -999.0, 'one_shot_yaw': -999.0}
+        # Save the first post maneuver / post ACQ attitude error sample for each AOATTER axis
+        pm_att_err = {}
+        for ax in [1, 2, 3]:
+            msid = aux_msidset['aoatter{}'.format(ax)]
+            ok = ((msid.times >= DateTime(guide_start).secs + 1.1)
+                  & (msid.times < DateTime(guide_start).secs + 30))
+            if not np.any(ok):
+                raise ValueError("No AOATTER{} times for guide transition at {} for one shot".format(
+                        ax, guide_start))
+            pm_att_err[ax] = msid.vals[ok][0] * R2A
+        return {'one_shot':  np.sqrt(pm_att_err[2] ** 2
+                                     + pm_att_err[3] ** 2),
+                'one_shot_roll': pm_att_err[1],
+                'one_shot_pitch': pm_att_err[2],
+                'one_shot_yaw': pm_att_err[3]}
 
     @classmethod
     @import_ska
@@ -1503,7 +1511,9 @@ class Manvr(TlmEvent):
         Get maneuver events from telemetry.
         """
         # Auxiliary information
-        aux_msidset = fetch.Msidset(['aotarqt1', 'aotarqt2', 'aotarqt3', 'one_shot'], start, stop)
+        aux_msidset = fetch.Msidset(['aotarqt1', 'aotarqt2', 'aotarqt3',
+                                     'aoatter1', 'aoatter2', 'aoatter3'],
+                                    start, stop)
         states, event_msidset = cls.get_msids_states(start, stop)
         changes = _get_msid_changes(list(event_msidset.values()),
                                     sortmsids={'aofattmd': 1, 'aopcadmd': 2,
@@ -1543,9 +1553,8 @@ class Manvr(TlmEvent):
                          )
             event.update(manvr_attrs)
             event.update(cls.get_target_attitudes(event, aux_msidset))
-            event['one_shot'] = cls.get_one_shot(manvr_attrs['guide_start'],
-                                                 aux_msidset['one_shot'])
-
+            one_shot = cls.get_one_shot(manvr_attrs['guide_start'], aux_msidset)
+            event.update(one_shot)
             dwells = cls.get_dwells(event, sequence)
             event['foreign']['Dwell'] = dwells
             event['n_dwell'] = len(dwells)
--- a/kadi/events/models.py
+++ b/kadi/events/models.py
@@ -1470,7 +1470,7 @@ class Manvr(TlmEvent):
         """
         Define one_shot attribute (one-shot attitude update following maneuver)
         """
-        # No acq => guide transition so no one-shot defined.  Use 0.0 as a convenience.
+        # No acq => guide transition so no one-shot defined.  Use -99.0 as a convenience.
         if guide_start is None:
             return -99.0
 
--- a/kadi/events/models.py
+++ b/kadi/events/models.py
@@ -1183,6 +1183,9 @@ class Manvr(TlmEvent):
               stop_roll      Float                                  Stop roll angle after manvr
                   angle      Float                                         Maneuver angle (deg)
                one_shot      Float                            One shot attitude update (arcsec)
+          one_shot_roll      Float                       One shot attitude update roll (arcsec)
+         one_shot_pitch      Float                      One shot attitude update pitch (arcsec)
+           one_shot_yaw      Float                        One shot attitude update yaw (arcsec)
     ==================== ========== ============================================================
 
     ``n_acq``, ``n_guide``, and ``n_kalman``: these provide a count of the number of times
@@ -1207,6 +1210,9 @@ class Manvr(TlmEvent):
 
     ``one_shot``: one shot attitude update following maneuver.  This is -99.0 for maneuvers
         with no corresponding transition to NPM.
+
+    ``one_shot_roll``, ``one_shot_pitch``, and ``one_shot_yaw`` are the values of AOATTER1, 2, and 3
+        from samples after the guide transition.
     """
     _get_obsid_start_attr = 'stop'  # Attribute to use for getting event obsid
     event_msids = ['aofattmd', 'aopcadmd', 'aoacaseq', 'aopsacpr']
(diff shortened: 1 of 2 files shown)


kadi/events/models.py
```diff
@@ -1183,6 +1184,9 @@ class Manvr(TlmEvent):
               stop_roll      Float                                  Stop roll angle after manvr
                   angle      Float                                         Maneuver angle (deg)
                one_shot      Float                            One shot attitude update (arcsec)
+          one_shot_roll      Float                                     One shot roll   (arcsec)
```
taldcroft: Stay with the `One shot attitude update ..` description not `One shot ..`.

kadi/events/models.py
```diff
@@ -1207,6 +1211,11 @@ class Manvr(TlmEvent):
 
     ``one_shot``: one shot attitude update following maneuver.  This is -99.0 for maneuvers
         with no corresponding transition to NPM.
+
+    ``one_shot_roll``, ``one_shot_pitch``, and ``one_shot_yaw`` are from the same record as ``one_shot``
```
taldcroft: Line break after `record`.

kadi/events/models.py
```diff
@@ -1466,23 +1481,33 @@ def get_target_attitudes(cls, event, msidset):
 
     @classmethod
     @import_ska
-    def get_one_shot(cls, guide_start, one_shot_msid):
+    def get_one_shot(cls, guide_start, aux_msid):
```
taldcroft: Call this `aux_msidset` for consistency with general usage in kadi.

kadi/events/models.py
```diff
@@ -1466,23 +1481,33 @@ def get_target_attitudes(cls, event, msidset):
...
-        if i0 == 0 or i1 == 0 or i0 == n_times or i1 == n_times:
-            raise ValueError('Unable to find complete data for one-shot attitude value'
-                             ' {} {} {} {} {}'
-                             .format(i0, i1, t0, np.min(one_shot_msid.times),
-                                     np.max(one_shot_msid.times)))
-        return np.max(one_shot_msid.vals[i0:i1])
+        aoatter1 = aux_msid['aoatter1']
+        aoatter2 = aux_msid['aoatter2']
+        aoatter3 = aux_msid['aoatter3']
+        if len(aoatter2.times) == 0:
```
taldcroft: I think that this can never happen because of [this checking](https://github.com/sot/kadi/blob/f1b13436381285dbb7956d27528777dc745638d0/kadi/events/models.py#L1505).  But fair enough to still check.

kadi/events/models.py
```diff
@@ -1466,23 +1481,33 @@ def get_target_attitudes(cls, event, msidset):
...
+            raise ValueError("No AOATTER2 times for guide transition at {} for one shot".format(
+                    guide_start))
+        dtime = aoatter2.times - DateTime(guide_start).secs
```
taldcroft: It turns out that `aux_msidset` is not interpolated to common times so you cannot assume that they are correlated in index.  You probably need to loop through each one independently.

kadi/events/models.py
```diff
@@ -1466,23 +1481,33 @@ def get_target_attitudes(cls, event, msidset):
...
+        # Define the index of the one shot as the sample closes to the guide transition
+        # That guide_start from AOACASEQ should be coming from the same telemetry as AOATTER*
+        # so the closest timestamp should really be from the same timestamp
```
taldcroft: Since the `guide_start` telemetry has not been interpolated to the same time stamps as the aux msids, you can't make this assumption.  Instead just look for the first value with `guide_start <= time < guide_start + 30` (independently for each axis).

kadi/events/models.py
```diff
@@ -1531,9 +1558,9 @@ def get_events(cls, start, stop=None):
                          )
             event.update(manvr_attrs)
             event.update(cls.get_target_attitudes(event, aux_msidset))
-            event['one_shot'] = cls.get_one_shot(manvr_attrs['guide_start'],
-                                                 aux_msidset['one_shot'])
-
+            one_shot = cls.get_one_shot(manvr_attrs['guide_start'],
+                                        aux_msidset)
```
taldcroft: Line break no longer required.

kadi/events/models.py
```diff
@@ -22,6 +22,7 @@
 
 ZERO_DT = -1e-4
 MAX_GAP = 328.1  # Max gap (seconds) in telemetry for state intervals
+R2A = 206264.8
```
taldcroft: Add `# Convert from radians to arcsec`

kadi/events/models.py
```diff
@@ -1466,23 +1481,33 @@ def get_target_attitudes(cls, event, msidset):
...
+        aoatter2 = aux_msid['aoatter2']
+        aoatter3 = aux_msid['aoatter3']
+        if len(aoatter2.times) == 0:
+            raise ValueError("No AOATTER2 times for guide transition at {} for one shot".format(
+                    guide_start))
+        dtime = aoatter2.times - DateTime(guide_start).secs
+        # Define the index of the one shot as the sample closes to the guide transition
+        # That guide_start from AOACASEQ should be coming from the same telemetry as AOATTER*
+        # so the closest timestamp should really be from the same timestamp
+        one_shot_idx = np.argmin(np.abs(dtime))
```
taldcroft: **ALWAYS** use `idx` in this kind of situation, unless:
- The variable is used over a long context (i.e. more than 20-ish lines away)
- You need disambiguate two `idx` type variables in the same local context.

Using the longer version decreases readability for me, the reviewers because it adds visual noise to the code without improving understanding.

kadi/events/models.py
jeanconn: I think my problem is not that they aren't interpolated to common times (I still don't see how fetch could get different records from the same PCAD3ENG file), but that Msidset could mark and remove different samples as bad, right?

kadi/events/models.py
jeanconn: Sorry, I originally was comparing 2 different indices (the old code value and my new value).  This PR may not have been ready for prime time when I started it yesterday.

kadi/events/models.py
jeanconn: Might be useful (to me anyway) to add the (new?) Post Maneuver Attitude Error (PMAE) description/definition in here too.

kadi/events/models.py
jeanconn: Should I also make an effort to guarantee the sample from each axis is from the same time?

taldcroft: I thought about this but I don't think it is necessary.  It's not quite like quaternions where you just get the wrong answer.  Here we just want the first available sample in each axis.

jeanconn: Great.  

For a rainy day, it might be a nice feature of something in fetch to basically remake the table of telemetry (in this case, the pcad3eng one) with guaranteed shared times across the rows and no interpolation.

jeanconn:
I think I've executed all the requested changes here, but I'm not sure what this needs for testing. 

Also, I tried to work the steps in https://github.com/sot/kadi/blob/master/NOTES.add_event_type, but did not understand the different update time ranges (or how to select the appropriate time range) in https://github.com/sot/kadi/blob/master/NOTES.add_event_type#L38 and when I got to https://github.com/sot/kadi/blob/master/NOTES.add_event_type#L72 I was seeing diffs that looked undesired (looked like something (astropy.table?) compressing the doc table).   Wasn't sure if I should update the code/process or push back to @taldcroft .

kadi/events/models.py
```diff
@@ -1466,23 +1481,30 @@ def get_target_attitudes(cls, event, msidset):
...
+        # Save the first post maneuver / post ACQ attitude error sample for each AOATTER axis
+        pm_att_err = {}
+        for ax in [1, 2, 3]:
+            msid = aux_msidset['aoatter{}'.format(ax)]
+            ok = ((msid.times >= DateTime(guide_start).secs)
+                  & (msid.times < DateTime(guide_start).secs + 30))
+            if not np.any(ok):
+                raise ValueError("No AOATTER{} times for guide transition at {} for one shot".format(
+                        ax, guide_start))
+            pm_att_err[ax] = msid.vals[ok][0]
```
taldcroft: DRY: put the R2A scaling here.

kadi/events/models.py
```diff
@@ -1466,23 +1481,30 @@ def get_target_attitudes(cls, event, msidset):
...
-                             .format(i0, i1, t0, np.min(one_shot_msid.times),
-                                     np.max(one_shot_msid.times)))
-        return np.max(one_shot_msid.vals[i0:i1])
+            return {'one_shot': -99.0,
+                    'one_shot_roll': -999.0, 'one_shot_pitch': -999.0, 'one_shot_yaw': -999.0}
+        # Save the first post maneuver / post ACQ attitude error sample for each AOATTER axis
+        pm_att_err = {}
+        for ax in [1, 2, 3]:
+            msid = aux_msidset['aoatter{}'.format(ax)]
+            ok = ((msid.times >= DateTime(guide_start).secs)
```
taldcroft: I'm getting slightly squeamish about a possible off-by-one error when pushing right up against the `guide_start` boundary.  So I would suggest making this `guide_start + 1.1 seconds` to force using the next telemetry sample.

taldcroft: For testing you could plot the difference between current and new `one_shot` values for a representative sample (e.g. a year of one shots).  Hopefully those will all be less than a few arcsec, and that would be sufficient testing for merge.

taldcroft:
For the docs, change the `out_lines = dat.pformat()` call to `dat.pformat(max_lines=-1, max_width=-1)`

taldcroft:
And it has probably gotten to the point where you might as well squash down the commits.  I don't think the commit history has a lot of value now.

kadi/events/models.py
jeanconn: I'm fine with going with guide_start + 1.1, but I'm having trouble constructing a situation in my head where you could have a sample with a time greater than guide_start but still not be in GUIDE.  Do you think there are some cases where AOATTER takes more than one sample to update (if so, 1.1 secs seems like it wouldn't help)?  Or if the value of `guide_start` can be wrong (too early) that sounds like a different problem that might need addressing.  Otherwise, the AOATTER values are still from the same telemetry as AOACASEQ and are not interpolated.

taldcroft: I can't think of a credible case either, but I do think it is credible that I have missed some possibility.  Ergo just use the next sample which will be just fine for this purpose.

jeanconn:
Regarding my comment above, I'm still not sure about 

> Also, I tried to work the steps in https://github.com/sot/kadi/blob/master/NOTES.add_event_type, but did not understand the different update time ranges (or how to select the appropriate time range) in https://github.com/sot/kadi/blob/master/NOTES.add_event_type#L38

taldcroft:
About the time ranges, did the dates provided in the instructions not work?  I don't remember the details.

jeanconn:
The instructions say:

```
./update_events --start=1999:200 --stop=2001:001 --model=${ModelClassName}
# The start date might need to be a bit later, e.g. for Manvr use 1999:230 + 21 days (lookback
# time) so that all the event_msids are fully sampled.  Probably not needed for events
# that rely on only one event MSID.
./update_events --start=2001:001 --model=${ModelClassName}
```

So it is doing two time ranges without telling me why that's needed and I don't know how to determine that "all the event_msids are fully sampled".  No errors.

taldcroft:
The two time ranges is probably just to make sure it seems to be working OK without generating the whole table.  If you are getting a decent output then it's good to go.

jeanconn:
Regarding validation, in addition to adding the per axis values, this code makes the fundamental change that the one shot is defined as the RSS of the first available values of AOATTER2 and AOATTER3 after changing to GUIDE.  The previous code defined the one shot as the maximum value of the RSS "within" 10 seconds of the guide transition.  I note, from looking at, say the eng archive derived one shot for obsid 54592, that there are real differences in results from the old and new methods.  In this case, the attitude errors before GUIDE were larger than after

```
start = '2012:145:03:27:09.172'
telem = fetch.Msid("one_shot", DateTime(start).secs - 10, DateTime(start).secs + 10)
telem.plot(marker='.', linestyle="")
```
image

The values also increased after GUIDE began

```
telem = fetch.Msid("one_shot", DateTime(start).secs, DateTime(start).secs + 40)
telem.plot(marker='.', linestyle="")
```
image

The current kadi one_shot value is, as expected, 34.28 and the value from this PR is 19.4 arcsec.  This seems correct and reasonable to me.  Do we need to do anything else WRT to change the values of the historical one shots this way.

Here's a histogram of the differences between old and new one_shot values in kadi looking at the last 5 years of maneuvers.

one_shot_diff

taldcroft:
👍  !

PR #113, opened 2018-07-01
Changes to allow passing tests on Mac local

taldcroft:
Requires the update to `cmd_states` as well.

Passes tests on Mac and HEAD linux.


taldcroft:
This affects only testing and occweb access.  Merging now for installation at a convenient time.

PR #116, opened 2018-08-03
Add telemetry format and subformat states

taldcroft:
To do:

- [x] Test of subformat (pending data from Larry)


taldcroft:
Passing tests on Ska and Ska3.

PR #117, opened 2018-08-03
 Add VCDU to commands archive and Py3 update capability

taldcroft:
This also includes tweaks to `update_cmds.py` so that it generates cmds files using Ska3 (current /proj/sot/ska3/test 2018.08.02-a5c0a36) which pass tests.

### Testing

**Update_cmds with /proj/sot/ska3/test**
- [x] Pass tests using /proj/sot/ska3/test
- [x] Pass tests using /proj/sot/ska

**Update_cmds with /proj/sot/ska/flight**
- [x] Pass tests using /proj/sot/ska3/test
- [x] Pass tests using /proj/sot/ska

### Deployment

- Wait for subsequent PR to make updates hourly.

### DIFF
--- a/kadi/update_cmds.py
+++ b/kadi/update_cmds.py
@@ -5,6 +5,7 @@ import difflib
 
 import numpy as np
 import tables
+import tables3_api
 from six.moves import cPickle as pickle
 import six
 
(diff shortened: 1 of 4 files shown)


kadi/update_cmds.py
```diff
@@ -399,7 +402,7 @@ def main(args=None):
 
     if pars_dict.n_updated > 0:
         with open(pars_dict_path, 'wb') as fh:
-            pickle.dump(pars_dict, fh, protocol=-1)
+            pickle.dump(pars_dict, fh, protocol=2)
```
jeanconn: For our Py2, was protocol=-1 (highest allowed, right?) equivalent to protocol=2?

taldcroft: Correct.  For the record:
https://docs.python.org/2/library/pickle.html#data-stream-format

jeanconn: Yes, that's what I figured.  So protocol=2 is a no-op for this if run on Py2 and locks it at a Py2-compatible protocol if run on Py3.

PR #118, opened 2018-08-06
Break events and commands processing into separate task_schedules

taldcroft:
This is to allow commands processing and ftp to GRETA to happen more frequently, say once per hour.

The only non-obvious content change is adding the `--occ` flag to the command for the OCC config files.  This flag has a default of True if the user is SOT, but better to just be explicit.


jeanconn:
Looks good to me.  I'd nit-pick that task_schedule_occ_cmds.cfg could use a new "Subject".

taldcroft:
Fixed the subject in 7cb5571.

PR #122, opened 2018-08-23
WSVIDALLDN should set ccd_count = 0

jzuhone:
The ACIS command `WSVIDALLDN` powers down all the video boards, and as such should set `ccd_count = 0`.

This is essentially identical to what is implemented in sot/cmd_states#39.


taldcroft:
In this case it should be straightforward to add a test.  Code looks fine.

taldcroft:
Ping @jzuhone - for the record, waiting for either a new unit test or at least a functional description of local testing that demonstrates this works.

jzuhone:
@taldcroft right--sorry. got sidetracked by some other stuff, will get back to this soon hopefully.

PR #131, opened 2019-06-14
Add backstop handling to kadi.commands

### LINKED ISSUE #130: How to build dynamic states from continuity and oflsdir backstop?
What is the recommended path to work from dynamic states to run thermal models?  

I assume that we want something similar to the cmd_states approach where we get:

   * continuity initial state
   * continuity states or commands from some time before the last available telemetry
   * (telemetry for thermistor data)
   * backstop cmds

And then from the initial state and the commands from the commands table and the commands in the under-evaluation backstop we make states.  I just don't immediately see how to extract commands from the backstop in products and combine those with the commands from kadi.commands.commands.get_cmds and feed everything to kadi.commands.get_states .

Specifically I was wondering about moving starcheck to dynamic states as part of the fix for 

https://github.com/sot/starcheck/issues/219

taldcroft:
Closes #130

### DIFF
--- a/kadi/commands/commands.py
+++ b/kadi/commands/commands.py
@@ -129,12 +129,14 @@ def get_cmds_from_backstop(backstop, remove_starcat=True):
 
     n_bs = len(bs)
     out = {}
-    out['idx'] = np.zeros(n_bs, dtype=np.uint16)
+    # Set idx to max (2**16 -1) so it does not match any real idx
+    out['idx'] = np.full(n_bs, fill_value=65535, dtype=np.uint16)
     out['date'] = np.chararray.encode(bs['date'])
     out['type'] = np.chararray.encode(bs['type'])
     out['tlmsid'] = np.chararray.encode(bs['tlmsid'])
     out['scs'] = bs['scs'].astype(np.uint8)
     out['step'] = bs['step'].astype(np.uint16)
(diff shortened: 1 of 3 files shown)


kadi/commands/commands.py
```diff
@@ -1,17 +1,18 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import tables
 import tables3_api
+from pathlib import Path
```
jeanconn: I think kadi in general is still Python 2 compatible?  And I've forgotten the transition plan and/or requirements.

taldcroft: Good point.  So this will drive the first 4.x version of kadi.  This change will not change the backing django database so that the current Py2 compatible 3.x version will continue working.

About the kadi web server, that's a big can of worms that I need to reopen.  IIRC the transition plans and requirements have not been formulated or written down, so you did not forget them.

jeanconn:
Since this was not working for me from the [starcheck example](https://github.com/sot/starcheck/pull/319#issuecomment-505546129), I thought that needed clearing up before running with this.

kadi/commands/commands.py
jeanconn: For kadi versioning, was your plan 4.17 or some such?

jeanconn:
Works for me now as expected from starcheck with the 31491de fix.

PR #136, opened 2019-08-30
 Set env vars to work on web-kadi / web-kadi-test servers

taldcroft:
Tested servers with these updates in place on web-kadi-test.


taldcroft:
But of course it appears what is actually running on the web server is not the 3.12.2 tagged release  (aka `web-kadi` branch) despite the version being set as such.  Sigh, will need to clean that up.

PR #140, opened 2019-09-15
Fix PEP8 and flake8 and clean Py2-only cruft

taldcroft:


kadi/cmds/cmds.py
```diff
@@ -1,6 +1,5 @@
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
 import tables
-import tables3_api
```
taldcroft: This is a Py2/Py3 compatibility module for PyTables that is no longer neeed.

kadi/commands/tests/test_states.py
```diff
@@ -13,7 +13,7 @@
 try:
     fetch.get_time_range('dp_pitch')
     HAS_PITCH = True
-except:
+except Exception:
     HAS_PITCH = False
```
javierggt: do we have a policy on exceptions? I know that some linters would complain about this exception being too generic.

taldcroft: Yes: https://occweb.cfa.harvard.edu/twiki/bin/view/Aspect/SkaPython#Catching_exceptions

@javierggt - you can't see that yet so I have emailed it to you.

taldcroft:
About `# noqa`, that's sort of standard I think we'll get used to the limited situations where it is used (basically import * and the rare instance of line > 100 chars that should be left).  

But agreed that the one non-obvious case of the `table` import by the decorator, that should be commented.  I'll do that.

taldcroft:
Good point on the additional scope of dropping support for Py2.  So I added that to the title.  That's where we have to be a bit pragmatic.  :smile:.

jeanconn:
Though I thought we dropped Py2 support at some point in the past when this became a version 4.x module; the tables3_api import might have been more Py2 cruft than Py2 support.

javierggt:
yes, I'm all in for dropping support for py2. Still :)

kadi/commands/tests/test_states.py
javierggt: ok, I can't say anything about whether the exceptions caught are the right ones, so I just mentioned it.

taldcroft: The truth is that I don't know offhand the exception that will get raised.  I can find it for one particular case, and maybe that might be fine idea.  But at the end there is a bit of pragmatism that the code inside tests need not be perfect and so I just did a programmatic replace of `except:` with `except Exception:`.

taldcroft:
Correct, Py2 support was dropped previously.  Title changed again!  "Fix PEP8 and flake8 and clean Py2-only cruft"

PR #147, opened 2019-12-04
Update mid-manvr test to use column comparisons and isclose

jeanconn:
Update mid-manvr test to use column comparisons and isclose

### DIFF
--- a/kadi/commands/tests/test_states.py
+++ b/kadi/commands/tests/test_states.py
@@ -1211,8 +1211,7 @@ def test_get_pitch_from_mid_maneuver():
     """
     start = '2019:039:14:16:25.002'  # Mid-maneuver
     stop = '2019:039:16:00:00.000'
-    exp = Table.read(
-          ['      datestart              datestop           pitch     pcad_mode',
+    exp = ['      datestart              datestop           pitch     pcad_mode',
            '--------------------- --------------------- ------------- ---------',
            '2019:039:14:16:25.002 2019:039:14:16:54.364 63.2696565838      NMAN',
            '2019:039:14:16:54.364 2019:039:14:22:01.825 83.0388345752      NMAN',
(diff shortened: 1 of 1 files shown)


kadi/commands/tests/test_states.py
```diff
@@ -1211,17 +1211,23 @@ def test_get_pitch_from_mid_maneuver():
     """
     start = '2019:039:14:16:25.002'  # Mid-maneuver
     stop = '2019:039:16:00:00.000'
-    exp = ['      datestart              datestop           pitch     pcad_mode',
+    exp = Table.read(
+          ['      datestart              datestop           pitch     pcad_mode',
```
jeanconn: I think this change violates flake8 because I tried to leave the columns where they were.  Up to @taldcroft

jeanconn:
I just tossed this here because the test is failing in my tests in a shiny ska3-dev:
```        del sts['trans_keys']

>       assert sts.pformat(max_width=-1, max_lines=-1) == exp
E       AssertionError: assert ['      dates...   NMAN', ...] == ['      dates...   NMAN', ...]
E         At index 0 diff: '      datestart              datestop             pitch        pcad_mode' != '      datestart              datestop           pitch     pcad_mode'
E         Use -v to get the full diff

kadi/commands/tests/test_states.py:1227: AssertionError
```

kadi/commands/tests/test_states.py
taldcroft: Better to leave the original `exp = [' .... ']` code and add `exp = Table.read(exp, format='ascii')`.  It's preferred not to have giant args in a function call.

taldcroft:
I'm concerned about the rather large matching tolerance of 0.01 deg.  That is much more than a numerical issue, so I'd like to see exactly what the diffs are and investigate the source of discrepancy.

jeanconn:
What tolerance would you like?  I thought the test should be verifying pitch was within reasonable tolerances for use in the thermal model so just picked a tolerances as such.  If you'd just like it to the 9 printed decimals you had that is fine too.

taldcroft:
This and many other Ska regression tests also serve the important purpose of validating that we can change numpy version or other core packages without changing downstream program behavior *at all*.  This is how we can justify the `shiny` branch.  So ideally we would like the answers to be the same to binary precision (around 2**-52 or 1e-15 in the mantissa), but in practice that is likely to fail so we can check at something like atol of 1e-8 here.

The key thing is to write tests that are as strict as possible and so we have visibility into regressions (often unrelated to the package itself being tested) and can then understand the issue and whether it might impact other code where testing is not as complete.

taldcroft:
BTW was the change just a slight formatting diff in the table output or a real numerical change?

jeanconn:
I thought it was a formatting diff, but of course not easy to see with the previous test.

PR #143, opened 2019-10-09
Py3 server with  + event/cmd cron jobs processing with django 3.0

taldcroft:
This is based on the scm_version branch.  Features:

- Make the django server run with Python 3 and django 3.0
- Make the events update work on Python 3
- Add `obsid` as a standard field for telemetry-based events
- Change name of event database file from `events.db3` to `events3.db3`.  New file and schema are not compatible with previous version.
- Flake8 changes

## Skare3 package requirements

In order to pass all documented tests these package versions or branches are required:

- django 3.0.1
- ska_helpers `run-info` branch (https://github.com/sot/ska_helpers/pull/6)
- find_attitude `py3-server` branch (https://github.com/sot/find_attitude/pull/7)
- mica `py3-server` branch (https://github.com/sot/mica/pull/213)
- task_schedule `modernize` branch (https://github.com/sot/task_schedule/pull/15)
- watch_cron_logs `modernize` branch (https://github.com/sot/watch_cron_logs/pull/6)
- ska_sync (update to sync events3.db3, https://github.com/sot/ska_sync/pull/15)

### Create test environment

Roughly speaking, on kady:
```
$ which conda
/export/tom/miniconda3/envs/ska3-django3/bin/conda
$ conda --version
conda 4.3.21
$ conda create -n ska3-django3 --clone=ska3
$ source activate ska3-django3
$ conda uninstall --force ska3-flight ska3-core django kadi

# Original ska3 env doesn't have perl (by mistake) so fix that here.
# Without --force it wanted to re-install old versions of kadi, django, gcc and 
# other random stuff
$ conda install --force -c $ska3conda task_schedule
$ conda install --force -c $ska3conda -c $ska3conda/core-pkg-repo perl perl-core-deps perl-ska-classic

$ pip install django  # version 3.0.1
$ pip install kadi  # From git repo
Successfully installed kadi-4.18.2.dev45+g9478c71
$ pip install find_attitude  # From git repo; version 3.3.dev2+gbe4485b (py3-server branch)
$ pip install mica  # From git repo py3-server branch
Successfully installed mica-4.20.dev3+gf5136d6

$ cd ~/git/task_schedule  # modernize branch
$ make install
$ cd ~/git/watch_cron_logs  # modernize branch
$ make install 

# Data links
$ ls -l /export/tom/miniconda3/envs/ska3-django3/data
total 4
[...]

PR #150, opened 2019-12-16
Changes to use scm_setuptools

javierggt:
Since I do not know how this affects backwards compatibility, I tried to leave the interface mostly unchanged. I think each package might have to be fixed to cleanup its interface.

In the case of kadi, the version is specified in various ways:
- `__version__`
- package_version
- `__git_version__`
- VERSION
- version

Right now all come from the same source.

**I expect the changes in eng_archive to be similar**


taldcroft:
I suspect that much or all of the complexity in the current versioning output is really not used anywhere.  E.g. https://github.com/search?q=org%3Asot+git_version&type=Code shows 9 places *defining* `git_version` but no configured code actually using it.  The problem this was solving is being able to relate the version of imported code back to a specific commit or tag of the package.  With setuptools_scm we can now do this reliably, without hacks like writing a GIT_VERSION file.

Which is a long way of saying that I propose just chopping the `version.py` file entirely and get rid of this historical cruft.  So long as all testing passes, then we're good to go.  Keep this PR open for a little while for reference, but basically I think we can say we are making a back-incompatible API change and if code needs fixing then we do it (fix and forward).

taldcroft:
Have not tested but this all looks about right to me.

jeanconn:
A nit, but probably also need an update in `doc/conf.py` .

jeanconn:
And just searching for other uses of `kadi.version` it looks like there's one in find_attitude, so that probably needs a new issue

https://github.com/sot/find_attitude/blob/cefd1923ad2ddde2496f98943f6fb1e2b504866a/find_attitude/web/views.py#L5

if I'm understanding correctly that `kadi.version` will just not exist.

javierggt:
Good catch. I will look for similar issues all over

javierggt:
Looking for similar cases, I found a few, including:
- starcheck/src/starcheck.pl (it is in a big string that I suppose gets run... ??)
- starcheck/calc_ccd_temps.py

jeanconn:
Yes, when you/we do starcheck we'll need to clean up some of its internal use of `starcheck.version`.  And yes, that string in starcheck.pl is used as`Inline::Python` block (though I think it is a separate starcheck issue to pull out more of the Python into files that can be appropriately flake8 checked and such instead of the string blocks).

jeanconn:
I think this one is just superseded by #143

PR #152, opened 2020-01-03
Add ACIS raw-mode SI modes to ACIS states

taldcroft:
The modern version of https://github.com/sot/cmd_states/pull/24.

Fixes https://github.com/sot/cmd_states/issues/23.


jeanconn:
I think this is non-controversial and can go without additional review.

PR #161, opened 2020-03-15
Add acisfp_setpoint state

taldcroft:
## Description

This adds a new state to track the ACIS FP temperature based on commands of the form `WSFTNEG*`.

This could have been done with far less new code by just adding a new `elif` block to the existing `ACISTransition` class.  However, the first `WSFTNEG` command in the commands table is in 2003, so querying states before then was crashing.  The current implementation puts in place a default of -121.0.

## Testing

- [x] Functional testing: new unit test

### DIFF
--- a/docs/commands_states.rst
+++ b/docs/commands_states.rst
@@ -509,8 +509,8 @@ classes which affect the keys.
 
 .. Run kadi.commands.states.print_state_keys_transition_classes_docs() to generate this list.
 
-``acisfp_temp``
-  - :class:`~kadi.commands.states.ACISFP_TempTransition`
+``acisfp_setpoint``
+  - :class:`~kadi.commands.states.ACISFP_SetPointTransition`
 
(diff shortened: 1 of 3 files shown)


jzuhone:
My quick comment here (before I do more testing) is that I would like to see a name change to reflect the fact that this is a setpoint change. So:

``acisfp_temp`` --> ``acisfp_setpt``
``ACISFP_TempTransition`` --> ``ACISFP_SetPointChange``

or something similar

jzuhone:
@taldcroft aside from the nit above, I'm generally good with this. I tried it out and it works fine. 

The one issue is that we did this during some CAPs in 2018, maybe about 3 times I think. I would have to dig, but I could probably get the times for those. What would be the best way to incorporate those?

taldcroft:
@jzuhone - yes, I had independently thought about calling this setpoint not temp.  This is done.

taldcroft:
> The one issue is that we did this during some CAPs in 2018, maybe about 3 times I think. I would have to dig, but I could probably get the times for those. What would be the best way to incorporate those?

If you can supply us with all non-load times and the exact command names (TLMSID value) we can fairly easily put this in using the existing machinery in Chandra.cmd_states.

PR #163, opened 2020-04-23
Add inclusive_stop kwarg to get_cmds

taldcroft:
## Description

This will make code related to RLTT a bit cleaner because in that case one needs to get commands up to and *including* the RLTT stop time.

This should be used in https://github.com/sot/starcheck/pull/341/.

## Testing

- [x] Functional testing
  - Added new unit test

### DIFF
--- a/kadi/commands/commands.py
+++ b/kadi/commands/commands.py
@@ -188,22 +188,20 @@ def _find(start=None, stop=None, inclusive_stop=False, **kwargs):
 
     :returns: astropy Table of commands
     """
-    date = kwargs.pop('date', None)
-    if date:
-        start = DateTime(date).date  # clip resolution to nearest msec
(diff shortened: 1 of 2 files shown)


kadi/commands/commands.py
```diff
@@ -184,14 +192,19 @@ def _find(start=None, stop=None, **kwargs):
     if date:
         start = DateTime(date).date  # clip resolution to nearest msec
         stop = DateTime(start) + 0.001 / 86400  # exactly 1 msec later
+        inclusive_stop = False
```
jeanconn, line 195: This makes sense to work with the logic below for inclusive_stop, but would there be value in just using some kind of "==" somewhere for the special case instead of generalizing?

taldcroft: Yup, done.

PR #190, opened 2020-08-19
Migrate cmds.h5 idx column from 16 to 32 bits

### LINKED ISSUE #189: failure when deriving states from backstop commands on shiny
the following script (run using shiny):

```python
from kadi import commands
from kadi.commands import states

# Get commands from JUN2720 from backstop
backstop_file = "/data/acis/LoadReviews/2020/JUL2720/oflsb/CR208_1806.backstop"
cmds = commands.get_cmds_from_backstop(backstop_file)

t = states.get_states(cmds=cmds, merge_identical=True)
```

produces the following traceback:

```pycon
  File "test_bs_cmds.py", line 8, in <module>
    t = states.get_states(cmds=cmds, merge_identical=True)
  File "/Users/jzuhone/miniconda3/envs/ska-shiny/lib/python3.8/site-packages/kadi/commands/states.py", line 1253, in get_states
    transitions = get_transitions_list(cmds, state_keys, start, stop, continuity)
  File "/Users/jzuhone/miniconda3/envs/ska-shiny/lib/python3.8/site-packages/kadi/commands/states.py", line 1128, in get_transitions_list
    transition_class.set_transitions(transitions_dict, cmds, start, stop)
  File "/Users/jzuhone/miniconda3/envs/ska-shiny/lib/python3.8/site-packages/kadi/commands/states.py", line 290, in set_transitions
    transitions_dict[date][name] = params[param_key]
KeyError: 'id'
```

Not sure if this is a bug or if I did something incorrect.

taldcroft:
## Description

The `idx` column in the `cmds.h5` file was unfortunately sized at unsigned 16-bit int (65536). This identifies unique command parameters. The initial estimate that this would be more than enough proved to be faulty and the number exceeded 65536 with the AUG1720 loads.

Key elements are:
- Changed from uint16 to int32 for the `idx` column.
- Added code to ensure that `idx=65536` is never used. This ensures that the current ska3-flight kadi version 5.2.0 runs correctly (in particular `get_cmds_from_backstop`) with the new files.

This PR includes a new utility script that was used to migrate the `cmds.h5` as of August 8 (which did not suffer from the overflow issue) to a new version. The script contains checks that the new and previous commands are identical except for the dtype of `idx`.

With this PR that `cmds.h5` and `cmds.pkl` files were updated to the present using `python -m kadi.update_cmds` from the git repo on this branch. They have been installed to `/proj/sot/ska/data/kadi`.

The footprint of the fix is small and no current flight code is impacted.

## Testing

- [x] Passes unit tests on linux (HEAD flight) with one workaround to ignore an expected failure in dtype comparison
- [x] Passes units tests on linux (HEAD shiny)
- [x] Functional testing

### Functional testing

**Original problem report**

```
from kadi import commands
from kadi.commands import states

# Get commands from JUN2720 from backstop
backstop_file = "/data/acis/LoadReviews/2020/JUL2720/oflsb/CR208_1806.backstop"
cmds = commands.get_cmds_from_backstop(backstop_file)

t = states.get_states(cmds=cmds, merge_identical=True)
```
This produces the expected output on HEAD (kady) and GRETA (cheru).

**Starcheck**

- Ran starcheck on HEAD on the AUG2420A loads. No process problems and gave the expected output.
- Ran starcheck on HEAD on the AUG1020A loads. Outputs identical to flight except for small propagation changes.

**File contents**

Manually examined the `cmds.pkl` file and confirmed that the `idx` value jumps as expected and no 65535 is found.

Fixes #189

### DIFF
--- a/kadi/commands/commands.py
+++ b/kadi/commands/commands.py
@@ -135,8 +135,8 @@ def get_cmds_from_backstop(backstop, remove_starcat=False):
 
     n_bs = len(bs)
     out = {}
-    # Set idx to max (2**32 -1) so it does not match any real idx
-    out['idx'] = np.full(n_bs, fill_value=4294967295, dtype=np.uint32)
+    # Set idx to -1 so it does not match any real idx
+    out['idx'] = np.full(n_bs, fill_value=-1, dtype=np.int32)
     out['date'] = np.chararray.encode(bs['date'])
     out['type'] = np.chararray.encode(bs['type'])
     out['tlmsid'] = np.chararray.encode(bs['tlmsid'])
--- a/kadi/commands/commands.py
+++ b/kadi/commands/commands.py
@@ -135,8 +135,8 @@ def get_cmds_from_backstop(backstop, remove_starcat=False):
 
     n_bs = len(bs)
     out = {}
-    # Set idx to max (2**16 -1) so it does not match any real idx
-    out['idx'] = np.full(n_bs, fill_value=65535, dtype=np.uint16)
+    # Set idx to max (2**32 -1) so it does not match any real idx
+    out['idx'] = np.full(n_bs, fill_value=4294967295, dtype=np.uint32)
     out['date'] = np.chararray.encode(bs['date'])
     out['type'] = np.chararray.encode(bs['type'])
     out['tlmsid'] = np.chararray.encode(bs['tlmsid'])
--- a/kadi/update_cmds.py
+++ b/kadi/update_cmds.py
@@ -440,7 +440,10 @@ def main(args=None):
 
     if pars_dict.n_updated > 0:
         with open(pars_dict_path, 'wb') as fh:
-            pickle.dump(pars_dict, fh, protocol=2)
+            # Dump as pickle, first converting pars_dict from UpdatedDict to
+            # plain dict. The n_updated attribute is not used outside of
+            # update_cmds.py. See historical note in NOTES.build for context.
+            pickle.dump(dict(pars_dict), fh, protocol=2)
             logger.info('Wrote {} pars_dict values ({} new) to {}'
                         .format(len(pars_dict), pars_dict.n_updated, pars_dict_path))
     else:
--- a/kadi/update_cmds.py
+++ b/kadi/update_cmds.py
@@ -19,7 +19,7 @@ from . import __version__
 
 MIN_MATCHING_BLOCK_SIZE = 500
 BACKSTOP_CACHE = {}
-CMDS_DTYPE = [('idx', np.uint16),
+CMDS_DTYPE = [('idx', np.uint32),
               ('date', '|S21'),
               ('type', '|S12'),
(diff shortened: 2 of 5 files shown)


kadi/commands/commands.py
```diff
@@ -135,8 +135,8 @@ def get_cmds_from_backstop(backstop, remove_starcat=False):
 
     n_bs = len(bs)
     out = {}
-    # Set idx to max (2**16 -1) so it does not match any real idx
-    out['idx'] = np.full(n_bs, fill_value=65535, dtype=np.uint16)
+    # Set idx to max (2**32 -1) so it does not match any real idx
+    out['idx'] = np.full(n_bs, fill_value=4294967295, dtype=np.uint32)
```
jeanconn, line 139: Thanks Tom!  Can you talk a bit more about how this fill value is used?   Trying to make sure I'm not missing a subtlety in my review.  I'm guessing the error on "no id" that we started to see came up because a command in the archive and a command or commands from the backstop ended up with the same dummy id?

taldcroft: It basically relates to the change below in `states.py`.  If the `idx` is not found in `REV_PARS_DICT` then it uses `cmd['params']` instead, which will always exist for commands read from backstop.

jeanconn: OK. I suppose the problem there is that I don't recall what REV_PARS_DICT is supposed to be etc and if that try/except is clever optimization or required logic etc.  Will puzzle for a while.

jeanconn, line 139: This looks makes sense and looks good.  At some point we might want to note somewhere that the 'idx' in a CommandTable from get_cmds_from_backstop doesn't look much like an index if all the records have the same value?

taldcroft: I changed the fill value to be -1, which is more obviously a non-valid value and is likely to provide a clue. It is worth noting that the string repr of a CommandTable does not include the `idx` column, meaning it is considered somewhat of a private attribute.

jeanconn: Makes sense.  I should have brought that up as a question yesterday (the "can we spare a bit for the sign and use -1" question or some such).

kadi/update_cmds.py
```diff
@@ -299,7 +299,12 @@ def get_idx_cmds(cmds, pars_dict):
         try:
             par_idx = pars_dict[pars_tup]
         except KeyError:
-            par_idx = len(pars_dict)
+            # Along with transition to 32-bit idx in #190, ensure that idx=65535
+            # never gets used. Prior to #190 this value was being used by
+            # get_cmds_from_backstop() assuming that it will never occur as a
+            # key in the pars_dict. Adding 65536 allows older versions to work
+            # with the new cmds.pkl pars_dict.
+            par_idx = len(pars_dict) + 65536
```
jeanconn, line 307: This looks fine but looks like it could cause problems if you are trying to rebuild from scratch?  Would need to poke around more to see how that case is handled.

PR #202, opened 2021-04-14
Make sure django is loaded if it is not already loaded

### LINKED ISSUE #201: kadi and subprocesses
We have discussed this multiple times. If we import kadi in a process, and then spawn a child process that also imports kadi, there is an exception in the child process.

Since it comes up often, I figured we would like to have a simple script to replicate the behavior to at document it at the very least, and perhaps we can fix it.

I created [a script in this gist](https://gist.github.com/javierggt/07850df0b76f07fc99a5b38f99c35813). It gives the time range of an OBSID given in the command line:
```
(ska3-flight-2021.2) javierg kadi $ python test_subprocess.py 18308
time range from OBSID: 2018:047:00:45:23.065-2018:047:03:36:33.565
```

One can pass the optional `--import` option, in which case it fails.

```
(ska3-flight-2021.2) javierg kadi $ python test_subprocess.py 18308 --import
Error getting OBS time range module 'kadi.events' has no attribute 'manvrs'
time range from OBSID: None-None
  File "test_subprocess.py", line 74, in <module>
    main()
  File "test_subprocess.py", line 71, in main
    start, stop = time_range_from_obsid(args.obsid)
  File "test_subprocess.py", line 54, in time_range_from_obsid
    raise Exception(f'OBSID time-range error: OBSID {obsid} not found in kadi')
Exception: OBSID time-range error: OBSID 18308 not found in kadi
```

javierggt:
## Description

This PR would Fix #201.

The way it is currently set up, `kadi` loads django if the DJANGO_SETTINGS_MODULE is not defined. The assumption is that, because django sets this variable during loading, if it is defined it means django is loaded. This fails when a child process is spawned, because the variable will be defined for children processes, but the children processes do no inherit the python module.

I added another condition for loading django. Namely, that the `manvrs` variable is not defined within `kadi.events`.

Possible things to do:
- remove the check for DJANGO_SETTINGS_MODULE
- does this check need to be made elsewhere?
- add a unit test case with the gist from #201 (or some other test involving spawned processes).

## Testing

- [x] Passes unit tests on MacOS. (well, it actually fails, but already failed before this change)
- [x] Functional testing. Used the gist linked to in issue #201

Fixes #

### DIFF
--- a/kadi/events/__init__.py
+++ b/kadi/events/__init__.py
@@ -53,7 +53,7 @@ import django
 # Jupyter notebook: SynchronousOnlyOperation: You cannot call this from an async
 # context. See: https://stackoverflow.com/questions/59119396
 
-if 'DJANGO_SETTINGS_MODULE' not in os.environ:
+if 'DJANGO_SETTINGS_MODULE' not in os.environ or 'manvrs' not in locals():
     os.environ['DJANGO_SETTINGS_MODULE'] = 'kadi.settings'
     os.environ["DJANGO_ALLOW_ASYNC_UNSAFE"] = "true"
     django.setup()


jeanconn:
I think this PR has been lingering and might be useful. @taldcroft did you have thoughts for more testing?  Largely I'm just wondering which open or merged PRs should really just get in to a ska3-matlab without much fuss.

taldcroft (approved):

PR #203, opened 2021-04-27
changed ALLOWED_HOSTS to include all host names we use.

javierggt:
## Description

This PR changes the ALLOWED_HOSTS variable in Django settings so it includes all our common hostnames. Perhaps we should consider going over the configuration.

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing on web-kadi-test server


taldcroft (approved):
LGTM

PR #205, opened 2021-09-15
Grating move duration in commanded states

taldcroft:
## Description

Per discussion on slack ("looking at the commands associated with the letg state key" in ska-dev on Aug 12), add a new intermediate state to the grating states to indicate the movement time between INSR and RETR.

Subsequent discussion revealed no user concerns about making this API change.

## Testing

- [x] Passes unit tests on MacOS, linux, Windows (at least one required)
- [N/A] Functional testing

### DIFF
--- a/kadi/commands/states.py
+++ b/kadi/commands/states.py
@@ -439,7 +439,7 @@ class HETG_INSR_Transition(MechMove):
     state_keys = ['letg', 'hetg', 'grating']
     transition_key = ['hetg', 'grating']
     transition_val = ['INSR', 'HETG']
-    move_duration = 203 * u.s
+    move_duration = 157 * u.s
 
 
 class HETG_RETR_Transition(MechMove):
@@ -448,7 +448,7 @@ class HETG_RETR_Transition(MechMove):
     state_keys = ['letg', 'hetg', 'grating']
     transition_key = ['hetg', 'grating']
     transition_val = ['RETR', 'NONE']
-    move_duration = 203 * u.s
+    move_duration = 153 * u.s
 
 
 class LETG_INSR_Transition(MechMove):
--- a/kadi/commands/states.py
+++ b/kadi/commands/states.py
@@ -12,8 +12,10 @@ import inspect
 import numpy as np
 
 from astropy.table import Table, Column
+import astropy.units as u
 
 from Chandra.Time import DateTime, date2secs, secs2date
+from cxotime import CxoTime
 import Chandra.Maneuver
 from Quaternion import Quat
 import Ska.Sun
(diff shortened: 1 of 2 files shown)


kadi/commands/states.py
```diff
@@ -378,36 +390,99 @@ class SubFormatSSR_Transition(FixedTransition):
 # Mech transitions
 ###################################################################
 
-class HETG_INSR_Transition(FixedTransition):
+class MechMove(FixedTransition):
+    """
+    Transitions for mech moves that having a non-zero duration.
```
javierggt, line 395: typo

kadi/commands/states.py
```diff
@@ -51,6 +52,17 @@
                       'vid_board')
 
 
+@contextlib.contextmanager
+def disable_grating_move_duration():
```
javierggt, line 56: It seems that this is used only in testing. Would this be required elsewhere? Can the reference data in testing be changed instead of using this function?

I am assuming this does not matter, but it changes a "static" class class member. There is not much threaded code, I suppose, but something like this could impact an instance running in another thread, right?

javierggt (changes requested):
the only request is the typo :-)

The other is more of a question.

kadi/commands/states.py
```diff
@@ -378,36 +390,99 @@ class SubFormatSSR_Transition(FixedTransition):
...
 
 
-class LETG_INSR_Transition(FixedTransition):
+class LETG_INSR_Transition(MechMove):
     """LETG insertion"""
     command_attributes = {'tlmsid': '4OLETGIN'}
     state_keys = ['letg', 'hetg', 'grating']
     transition_key = ['letg', 'grating']
     transition_val = ['INSR', 'LETG']
+    move_duration = 203 * u.s
```
jeanconn, line 476: But is it?  Too soon.  I think for modeling purposes this is perfectly sufficient.

kadi/commands/states.py
jeanconn: My impression is that it is only used in testing here but the intent is that this would be a public method to allow a user to turn off the grating "move" times as desired.  And that this code is probably not thread safe.

kadi/commands/states.py
jeanconn: I fixed the typo.

kadi/commands/states.py
taldcroft: I originally wrote this just for testing, but decided it could just as well go as a public method in case someone discovers they need to have the legacy version without the move states. A rule of thumb from astropy is that whenever you make a change that breaks a bunch of your own unit tests, then beware!

About being thread safe, good point. I think there is minimal risk of a problem here, but for the sake of argument how might this functionality be accomplished in a thread safe way?

kadi/commands/states.py
taldcroft: This is the number that Ken Gage provided. I believe it represents the max duration of a move with 24 short moves (each 2 sec) + the nominal 151.7 (ish).

kadi/commands/states.py
javierggt: It is not trivial to guarantee thread safety, and I know it is not warranted in this case because we don't use threads in this case. I would have to think.

A class member is basically a global variable, and global variables lead to unexpected behavior, because you as a user are not thinking about them. That is why this is my knee-jerk reaction.

It is a bit better if `apply_move_duration` is a data member of an instance, and not of the class, because in this case any thread-related issues would happen if you use the same instance at the same time from two threads, which can be expected. However, you always use the module-level functions, so I can't really propose a simple alternative.

taldcroft: Understood. With that I propose you just approve and we go ahead with this. Looks like @jeanconn fixed the typo.

jeanconn (approved):

PR #207, opened 2021-09-15
Improve list view and fix old web-kadi issue

taldcroft:
## Description

Fix the old issue https://github.com/sot/web-kadi/issues/2. The problem was that the template expects the first field to be the primary key. This PR takes the approach of just changing the field order to make that the case for `Orbit` events.

It also addresses the request to reverse the time order. This is done for all event types, which seems reasonable to me.

Lastly, this PR adds localhost to the ALLOWED_HOSTS list because the local server doesn't work without that.

Note: this is based on eed97c1 (Merge pull request #204 from sot/note-bad-dump) which is the commit before #202 that breaks using the local server.

## Testing

- [x] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing

### Functional testing

- Ran a local server and verified that Orbit events can be selected from the list view and properly displayed in detail view.
- Confirmed that the sort order is reversed in all event types.

Fixes https://github.com/sot/web-kadi/issues/2

### DIFF
--- a/kadi/events/views.py
+++ b/kadi/events/views.py
@@ -200,6 +200,10 @@ Examples:
                   if field.name not in self.ignore_fields]
         field_names = [field.name for field in fields]
 
+        # Get index of primary key in field_names. This gets used in the
+        # template to make a reference for the detail page.
+        context['model_pk_col_index'] = field_names.index(self.model._meta.pk.name)
+
         root_url = '/kadi/events/{}/list'.format(context['model_name'])
         filter_ = context['filter']
         sort = context['sort']
--- a/kadi/events/views.py
+++ b/kadi/events/views.py
@@ -15,7 +15,7 @@ MODEL_NAMES = {m_class.__name__: m_name
 
 
 class BaseView(object):
-    reverse_sort = False  # Reverse the default sort by model primary key (e.g. CAPs)
+    reverse_sort = True  # Reverse the default sort by model primary key (e.g. CAPs)
 
     def get_sort(self):
         sort = self.request.GET.get('sort')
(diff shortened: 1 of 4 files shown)


javierggt:
It seems ok to me. My main question is about process.

Will this change require a DB migration? because you changed the order of fields in the DB. How do you suggest we do the test then? I guess we have to:
- replicate $SKA/data/kadi (currently the test server runs directly on $SKA/data/kadi)
- Update in test environment
- Apply migrations there
- test

taldcroft:
Good question about the DB and possible migration. I assumed that since it works with the existing DB that all is OK. But I didn't try adding new data. It's actually not too long to regenerate the entire events database. In the end there is just the events3.db3 file that we can copy where needed. I'll think more on it.

taldcroft:
@javierggt - given the good question about the mis-ordering between the database fields and the model definition fields, I decided to dig in a bit and figure out a better way. This now just keeps track of the primary key column index and uses that in the template to make the detail view link.

This has the unfortunate feature of using the "private" model `_meta` attribute, but this is already used in other bits and if this part of the Django API every changes we will certainly notice it in functional testing.

javierggt (approved):
Looks good to me.

jeanconn (approved):
Looks good to me too.  I assume the functional testing was re-done after the model_pk_col_index tweak.

kadi/events/views.py
```diff
@@ -200,6 +200,10 @@ def get_context_data(self, **kwargs):
                   if field.name not in self.ignore_fields]
         field_names = [field.name for field in fields]
 
+        # Get index of primary key in field_names. This gets used in the
+        # template to make a reference for the detail page.
+        context['model_pk_col_index'] = field_names.index(self.model._meta.pk.name)
```
jeanconn, line 205: I like the comment!

PR #212, opened 2021-11-29
Implement commands archive v2.0 (beta release)

### LINKED ISSUE #172: get_states docstring looks inconsistent
After 

https://github.com/sot/kadi/commit/cb11b2ee03f0041728c457db963bcf60499770b8

I think this docstring needs update

https://github.com/sot/kadi/blob/6a8931cceb56ad0ecf7e6dd2b594682dad51fa0f/kadi/commands/states.py#L1131

unless I'm misunderstanding.

taldcroft:
## Description

This PR implements the [command archive v2.0 concept](https://docs.google.com/document/d/1UZfcQvVdIBNb3XdJnO8Ra_d5Xt8x_FHLWgVHa2FvHyE/edit#).

This requires https://github.com/sot/parse_cm/pull/35.

## Testing

- [x] Passes all unit tests on MacOS @ 246ceb9
- [x] Functional testing

### Interface impacts

- New `pprint_like_backstop` method on `CommandTable`

#### V1 commands (currently the default)
- Add a `time` column to the CommandTable returned by `get_cmds`

#### V2 commands (beta)
- Remove the `timeline_id` column and add a `source` column to the CommandTable.
- Access network resources (Google sheets, OCCweb) by default
- Add LOAD_EVENT pseudo-commands containing observation information
- Add `get_observations` and `get_starcats` functions.
- CommandTable string columns like `date` and `source` are kept in encoded ASCII `bytes` in the table instead of being converted to unicode for improved performance. The astropy Table "unicode-sandwich" machinery should make this largely transparent to the user.

#### Star and fid identification [OK]

Confirm that with 1.5 arcsec star search box and 40 arcsec fid search box that stars and fids are identified for every observation in the archive. It appears there are a handful of failed IDs for fids before 2008. I suspect these are defects in the commands archive itself, but in any case this low-level of problems for old observations is OK.

```
def test_get_starcats_each_year(year):
    starcats = get_starcats(start=f'{year}:001', stop=f'{year+1}:001')
    assert len(starcats) > 2
    for starcat in starcats:
        # Make sure fids and stars are all ID'd
        ok = (starcat['type'] != 'MON') & (starcat['type'] != 'FID')
        if np.any(starcat['id'][ok] == -999):
            print('Bad stars', starcat.obsid)
        ok = starcat['type'] == 'FID'
        if np.any(starcat['id'][ok] == -999):
            print('Bad fids', starcat.obsid, starcat.sim_offset, starcat.detector)

for year in range(2003, 2023):
    print(f'{year=}')
    test_get_starcats_each_year(year)

year=2003
Bad fids 3778 -1177 ACIS-I
Bad fids 4000 -1177 ACIS-I
Bad fids 4912 -4 HRC-S
year=2004
Bad fids 4937 -4 HRC-S
Bad fids 60116 -4 HRC-S
year=2005
Bad fids 5587 1 HRC-I
year=2006
Bad fids 6723 -1 ACIS-I
Bad fids 6432 -4 HRC-S
Bad fids 58647 -4 HRC-S
year=2007
Bad fids 7763 -2444 ACIS-S
Bad fids 7525 -2444 ACIS-S
year=2008
year=2009
year=2010
year=2011
year=2012
year=2013
year=2014
year=2015
year=2016
year=2017
year=2018
year=2019
year=2020
year=2021
year=2022
```

#### Task schedule [OK]

On Mac with this PR installed to a test Ska3 env:
```
$ task_schedule3.pl -config kadi/task_schedule_cmds.cfg
```
This ran successfully and updated `$SKA/data/kadi/cmds2.{h5,pkl}` as expected.

#### Configuration [OK]

```
$ rm ~/.kadi/config/kadi.cfg
$ ipython
In [1]: from kadi import create_config_file
   ...: from kadi.commands import conf
   ...: print(f'{conf.default_lookback=}')
   ...: create_config_file(overwrite=True)
conf.default_lookback=30
INFO: The configuration file has been successfully written to /Users/aldcroft/.kadi/config/kadi.cfg [astropy.config.configuration]
(ska3) ➜  kadi git:(commands-archive) head ~/.kadi/config/kadi.cfg 
[commands]

## Default lookback for previous approved loads (days).
# default_lookback = 30

## Cache backstop downloads in the astropy cache. Should typically be False,
## but useful during development to avoid re-downloading backstops.
# cache_loads_in_astropy_cache = True

## Clean backstop loads (like APR1421B.pkl.gz) in the loads directory that are
```
Now edit the config file to set `default_lookback = 60` and re-run but with `overwrite=False`:
```
In [1]: from kadi import create_config_file
   ...: from kadi.commands import conf
   ...: print(f'{conf.default_lookback=}')
   ...: create_config_file(overwrite=False)
conf.default_lookback=60
WARNING: The configuration file already exists and seems to have been customized, so it has not been updated. Use overwrite=True if you really want to update it. [astropy.config.configuration]
Out[1]: False
```

## Tentative migration plan

See #216

Fixes #172

### DIFF
--- a/docs/commands_states.rst
+++ b/docs/commands_states.rst
@@ -197,7 +197,8 @@ Version 2 of the commands archive is currently provided as an experimental
 release to allow getting experience with the new interface.
 
 For details of the commands v2 archive, including important information about
-the content and timeliness of commands, please see:
+network access, the timeliness of commands, configuration and process details,
+please see:
 
 .. toctree::
    :maxdepth: 2
@@ -213,9 +214,10 @@ In order to use the v2 version do the following::
   >>> from kadi import commands
   >>> commands.conf.commands_archive_version = "2"  # must be the string "2" not int 2
 
-An alternative is to set the `KADI_COMMANDS_VERSION` environment variable to `2`.
-This will globally apply to all subsequent Python sessions that inherit this
-environment. For example from a linux/Mac bash command shell you can enter::
+An alternative is to set the ``KADI_COMMANDS_VERSION`` environment variable to
+``2``. This will globally apply to all subsequent Python sessions that inherit
+this environment. For example from a linux/Mac bash command shell you can
+enter::
 
   $ export KADI_COMMANDS_VERSION=2
 
--- a/docs/commands_states.rst
+++ b/docs/commands_states.rst
@@ -2,6 +2,7 @@
 .. |get_continuity| replace:: :func:`~kadi.commands.states.get_continuity`
 .. |get_states| replace:: :func:`~kadi.commands.states.get_states`
 .. |CommandTable| replace:: :class:`~kadi.commands.commands.CommandTable`
+.. _Chandra Command Events: https://docs.google.com/spreadsheets/d/19d6XqBhWoFjC-z1lS1nM6wLE_zjr4GYB1lOvrEGCbKQ
 
 Commands and states
 ===================
@@ -17,8 +18,9 @@ As of this release there are two versions of the commands archive:
 - `Commands archive v1`_ (flight): this is the current default version and
   relies on iFOT load segments and the Chandra.cmd_states timelines database to
   maintain the commands database.
-- `Commands archive v2`_ (experimental): this relies on the Chandra Command
-  Events sheet and OCCweb FOT mission planning approved load products to
+- `Commands archive v2`_ (beta): this provides improved timeliness during
+  anomalies and better team-wide communication of non-load spacecraft commanding.
+  This relies on the `Chandra Command Events`_ sheet and OCCweb FOT mission planning approved load products to
   maintain the commands database. V2 is planned to be promoted to the default
   and the legacy will be deprecated and then retired.
 
@@ -206,8 +208,8 @@ please see:
    commands_v2.rst
 
 
-Getting started
-^^^^^^^^^^^^^^^
+Getting commands
+^^^^^^^^^^^^^^^^
 
 In order to use the v2 version do the following::
 
@@ -324,6 +326,153 @@ command table.  Note that command rows without that parameter will have a ``None
       None
       None
   AONM2NPE
+
+.. _getting-observations:
+
+Getting observations
+^^^^^^^^^^^^^^^^^^^^
+
+The commands archive includes special ``LOAD_EVENT`` commands (like the RLTT) that
+contain information about observations in the loads. An "observation" is defined
+as a dwell in normal point mode following a maneuver and includes most OR's and
+ER's. These are most easily accessed via the
+:func:`~kadi.commands.observations.get_observations()` function. For example::
+
+    >>> from kadi.commands import get_observations
+    >>> obss = get_observations(obsid=26330)
+    >>> obss[0]
+    {'obsid': 26330,
+    'simpos': 73296,
+    'obs_stop': '2022:075:19:10:35.734',
+    'manvr_start': '2022:075:17:39:54.696',
+    'targ_att': (0.105773397, -0.727314387, -0.579818109, 0.351620152),
+    'npnt_enab': True,
+    'obs_start': '2022:075:17:57:55.985',
+    'prev_att': (0.113370245, -0.848288771, -0.329080853, 0.399072852),
+    'starcat_idx': 211630,
+    'source': 'MAR1422A'}
+
+Notice that the command always returns a list of observations, even for a query
+asking for a specific ObsID. The reason is that cases of multiple observations
+with the same ObsID are relatively common, in particular after SCS-107 stops the
+observing loads this will happen. The commands archive reflects the commands
+that ran on-board, and since ObsID updates are in the observing loads those
+commands no longer run after SCS-107.
+
+For example, ObsID 65526 was manually commanded after the HRC B-side anomaly and
+persisted for 64 distinct observations in the vehicle loads::
+
+    >>> obss = get_observations(obsid=65526)  # ObsID after HRC B-side anomaly
+    >>> len(obss)
+    64
+
+Getting all the observations covering years or more is reasonably fast,
+typically seconds the first query and then << 1 sec for subsequent queries.
+
+    >>> %time obss = get_observations(start='2020:001', stop='2022:001')
+    CPU times: user 28.4 ms, sys: 952 µs, total: 29.4 ms
+    Wall time: 28.7 ms
+
+For a large number of observations like this you may find it convenient to turn
+this into an astropy ``Table``::
+
+    >>> from astropy.table import Table
+    >>> obss = Table(obss)
+    >>> obss
+    <Table length=4018>
+    obsid simpos        obs_stop            manvr_start      ...       obs_start               prev_att [4]         starcat_idx  source
+    int64 int64          str21                 str21         ...         str21                   float64               int64      str8
+    ----- ------ --------------------- --------------------- ... --------------------- ---------------------------- ----------- --------
+    47575  75624 2020:001:06:51:13.985 2020:001:06:15:56.053 ... 2020:001:06:46:33.236   -0.51363412 .. 0.832139148      166529 DEC2319A
+    23000  75624 2020:001:19:18:15.914 2020:001:06:51:24.236 ... 2020:001:07:17:16.164  -0.475120775 .. 0.693367999      166530 DEC2319A
+    47574  75624 2020:001:19:48:58.024 2020:001:19:18:26.165 ... 2020:001:19:44:17.276  -0.221358674 .. 0.710485779      166531 DEC2319A
+    ...    ...                   ...                   ... ...                   ...                          ...         ...      ...
+    25803  75624 2021:365:13:38:31.497 2021:365:05:20:58.062 ... 2021:365:05:47:31.748 -0.488490169 .. 0.0261316575      170515 DEC3021A
+    26264  75624 2021:365:18:39:19.983 2021:365:13:38:41.748 ... 2021:365:14:08:20.234 -0.0658948803 .. 0.350032226      170516 DEC3021A
+    26247  75624 2022:001:05:21:28.604 2021:365:18:39:30.234 ... 2021:365:19:06:05.650  0.553670113 .. 0.0635036584      170517 DEC3021A
+
+Under the hood
+""""""""""""""
+
+The observation information is stored as ``LOAD_EVENT`` commands that can be viewed
+directly::
+
+    >>> from kadi.commands import get_cmds
+    >>> cmds = get_cmds('2022:001', '2022:002', type='LOAD_EVENT')
+    >>> print(cmds)
+            date            type    tlmsid scs step      time      source  vcdu params
+    --------------------- ---------- ------ --- ---- ------------- -------- ---- ------
+    2022:001:05:48:44.808 LOAD_EVENT    OBS   0    0 757403393.992 DEC3021A   -1    N/A
+    2022:001:09:42:05.439 LOAD_EVENT    OBS   0    0 757417394.623 DEC3021A   -1    N/A
+    2022:001:11:37:20.405 LOAD_EVENT    OBS   0    0 757424309.589 DEC3021A   -1    N/A
+    2022:001:15:03:39.654 LOAD_EVENT    OBS   0    0 757436688.838 DEC3021A   -1    N/A
+    2022:001:15:29:26.255 LOAD_EVENT    OBS   0    0 757438235.439 DEC3021A   -1    N/A
+    2022:001:17:33:53.255 LOAD_EVENT    OBS   0    0 757445702.439 DEC3021A   -1    N/A
+    >>> cmds[0]['params']
+    {'obsid': 45814,
+    'simpos': -99616,
+    'obs_stop': '2022:001:09:13:04.557',
+    'manvr_start': '2022:001:05:21:38.855',
+    'targ_att': (0.530730117, 0.556620885, 0.610704042, 0.188518716),
+    'npnt_enab': True,
+    'obs_start': '2022:001:05:48:44.808',
+    'prev_att': (-0.0743435142, -0.559183412, -0.804323901, 0.186681591),
+    'starcat_idx': 170518}
+
+As with :func:`~kadi.commands.commands.get_cmds` in the v2 archive, you can provide a
+``scenario`` keyword to :func:`~kadi.commands.observations.get_observations` to
+select a custom or ``'flight'`` scenario.
+
+.. _getting-star-catalogs:
+
+Getting star catalogs
+^^^^^^^^^^^^^^^^^^^^^
+
+The ACA star catalogs associated with observations can be retrieved using the
+:func:`~kadi.commands.observations.get_starcats()` function. For example::
+
+    >>> from kadi.commands import get_starcats
+    >>> acas = get_starcats(obsid=26330)
+    >>> acas[0]
+    <ACATable length=11>
+    slot  idx      id    type  sz    mag    maxmag   yang     zang    dim   res  halfw
+    int64 int64   int64   str3 str3 float64 float64 float64  float64  int64 int64 int64
+    ----- ----- --------- ---- ---- ------- ------- -------- -------- ----- ----- -----
+        0     1         2  FID  8x8    7.00    8.00  -773.14 -1862.22     1     1    25
+        1     2         4  FID  8x8    7.00    8.00  2140.38    46.50     1     1    25
+        2     3         5  FID  8x8    7.00    8.00 -1826.24    40.03     1     1    25
+        3     4 194257752  BOT  6x6    6.09    7.59 -2389.55 -1716.62    28     1   160
+        4     5 264114816  BOT  6x6    8.64   10.19    -2.30 -2430.32    28     1   160
+        5     6 194249696  BOT  6x6    8.76   10.27 -2129.92 -2447.14    28     1   160
+        6     7 263198168  BOT  6x6    8.81   10.33  -375.16  2416.16    28     1   160
+        0     8 263199776  ACQ  6x6   10.13   11.20  1385.78  1949.73     8     1    60
+        1     9 264113448  ACQ  6x6   10.27   11.20  -187.29 -1336.05     8     1    60
+        2    10 263201064  ACQ  6x6   10.44   11.20   529.26   324.30     8     1    60
+        7    11 263196576  ACQ  6x6   10.56   11.20 -1046.29   258.97    16     1   100
+
+This function can be used to get star catalogs over a range of time. Depending
+on your specific needs, you might want to specify ``set_ids=False`` in the call,
+which disables the time-consuming step of finding the star and fid ids by
+cross-matching with expected star / fid locations. For example, to find catalogs
+between 2015 and 2022 with MON windows commanded::
+
+    >>> %time acas = get_starcats(start='2015:001', stop='2022:001', set_ids=False)
+    CPU times: user 13.4 s, sys: 91.8 ms, total: 13.5 s
+    Wall time: 13.5 s
+
+    >>> len(acas)
+    13261
+    >>> acas_with_mon = [aca for aca in acas if 'MON' in aca['type']]
+    >>> len(acas_with_mon)
+    119
+    >>> acas_with_mon[0].obsid
+    17320
+
+.. Note::
+   The ``ACATable`` objects that are returned can be plotted but they
+   are not fully equivalent to the catalogs that ``proseco`` would return. The
+   CCD temperatures are set to -20 C and the ``.acqs`` and ``.guides`` attributes
+   are stubbed with empty tables.
 
 Chandra states and continuity
 ------------------------------
--- a/docs/commands_states.rst
+++ b/docs/commands_states.rst
@@ -8,9 +8,19 @@ Commands and states
 
 **Commands**
 
-The `Commands archive`_ is a table of every load command that has been run, or is currently
+The Commands archive is a table of every load command that has been run, or is currently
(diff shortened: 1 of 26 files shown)


jskrist:
What do you think about using sheet id unstead of spreadsheet id for scenarios?  That way there could be one spreadsheet used for an anomaly response, and different sheets for each scenario being considered (with one central document holding everything).

see [Google Sheets API Doc](https://developers.google.com/sheets/api/guides/concepts) for information about the differences.

taldcroft:
@jskrist - About sheet id, I'm a little worried about managing permissions. Right now the idea is that "flight" Commands Sheet is editable by only a small set of people, but viewable by the world. I think if you make another sheet (tab) then it will have the same edit permissions, but the idea is that anyone can make their own scenario sheet and edit.

jskrist:
@taldcroft, that makes sense.  I see the power and flexibility of allowing individuals to make scenarios in their own spreadsheets as they are doing analysis,  but it puts back a little of the "where is the latest data?" problem, that this central sheet was resolving.

taldcroft:
@jeanconn - any progress on review of this PR?

jeanconn:
Not really, I'll have more for today.

kadi.cfg
```diff
@@ -0,0 +1,30 @@
+[commands]
+# Default lookback for previous approved loads (days).
+# default_lookback = 30
```
jeanconn, line 3: Why is it the defaults are also commented out in the config?

kadi/commands/__init__.py
```diff
@@ -1,2 +1,61 @@
...
+        'KADI_COMMANDS_VERSION environment variable.'
+    )
+
+    cmd_events_flight_id = ConfigItem(
+        '19d6XqBhWoFjC-z1lS1nM6wLE_zjr4GYB1lOvrEGCbKQ',
+        'Google Sheet ID for command events (flight scenario).'
+    )
+
+    star_id_match_halfwidth = ConfigItem(
+        5,
```
jeanconn, line 46: For other work we've had ID_DIST_LIM at 1.5 I think.

kadi/tests/test_occweb.py
```diff
@@ -97,3 +96,44 @@ def test_get_fdb_major_events():
...
+@pytest.mark.skipif('not HAS_OCCWEB')
+@pytest.mark.parametrize('cache', [False, True])
+def test_get_occweb_dir_fail(cache):
+    """Test get_occweb_dir and get_occweb_page (which is called in the process)"""
+    path = Path('FOT/mission_planning/PRODUCTS/APPR_LOADS/FAIL-NOT-THERE')
+    with pytest.raises(requests.exceptions.HTTPError):
+        occweb.get_occweb_dir(path, cache=cache)
+
+
+@pytest.mark.skipif('not HAS_OCCWEB')
```
jeanconn, line 131: Should these just have `reason` and match the "modern preferred style"?

kadi/tests/test_occweb.py
```diff
@@ -7,16 +8,14 @@
 # test collection by pytest.  Note that this does not work with
 # paramiko 2.0.0.
 from paramiko import py3compat
+import requests
 py3compat.u('dirty hack')
```
jeanconn, line 12: Is this still needed?

kadi/occweb.py
```diff
@@ -11,29 +11,38 @@
...
         'ifot': '/occweb/web/webapps/ifot/ifot.php',
         }
 
+# Initialize 'kadi.occweb' logger.
+logger = logging.getLogger(__name__)
+
+
+def get_auth(username=None, password=None):
+    if username and password:
+        return (username, password)
```
jeanconn, line 41: Not entirely sure I understand this use case.

kadi/paths.py
```diff
@@ -1,22 +1,70 @@
...
+
+def PARS_DICT_PATH(version=None):
+    version = _version_str(version)
+    return DATA_DIR() / f'cmds{version}.pkl'
+
+
+def CMDS_DIR(scenario=None):
+    from kadi.commands import conf
+
+    # Special case for "flight" scenario. This is hardwired to $SKA/data/kadi
```
jeanconn, line 36: If hardwired and intended for HEAD flight, shouldn't it just be fully hardwired or warn if it isn't?

kadi/commands/command_sets.py
```diff
@@ -0,0 +1,263 @@
...
+    from parse_cm.csd import csd_cmd_gen
+
+    rts = 'SCS_CATEGORY,OBSERVING\n' + '\n'.join(args).upper()
+    cmds = csd_cmd_gen(rts, date=date)
+    return cmds
+
+
+def cmd_set_obsid(obs_id, date=None):
+    """Return a command set that initiates a maneuver to the given attitude
+    ``att``.
```
jeanconn, line 25: Docstring seems a copy-paste error.

kadi/commands/command_sets.py
```diff
@@ -0,0 +1,263 @@
...
+
+def cmd_set_rts(*args, date=None):
+    from parse_cm.csd import csd_cmd_gen
+
+    rts = 'SCS_CATEGORY,OBSERVING\n' + '\n'.join(args).upper()
+    cmds = csd_cmd_gen(rts, date=date)
+    return cmds
+
+
+def cmd_set_obsid(obs_id, date=None):
```
jeanconn, line 23: And are a bunch of these set to take dates they don't use to make the API simpler?  Would generic kwargs make sense at that point?

kadi/commands/command_sets.py
```diff
@@ -0,0 +1,263 @@
...
+    cmd, = cmd_set_command(*args, date=date)
+    cmd['type'] = 'NOT_RUN'
+    return (cmd,)
+
+
+def get_cmds_from_event(date, event, params_str):
+    r"""
+    Return a predefined cmd_set ``name`` generated with \*args.
+
+    :param name: cmd set name (manvr|scs107|nsm)
```
jeanconn, line 185: Not really clear on how these params line up with the args.

kadi/commands/__init__.py
taldcroft: I was a little bit worried (without any particular evidence) about getting failed matches going way back in time e.g. to 2002. I guess for the most part all those catalogs went through starcheck and got independent cross-matching at 1.5 arcsec from starcheck. I'm fine dropping this to 1.5 arcsec.

kadi/tests/test_occweb.py
taldcroft: Indeed, will fix!

kadi/tests/test_occweb.py
taldcroft: I took it out and test_occweb tests seems to pass. I'll push the commit and see if it flies in full integration testing.

kadi/occweb.py
taldcroft: This allows the user / calling function to explicitly pass in credentials instead of getting them automatically through other means. See e.g. `get_occweb_page`.

kadi.cfg
taldcroft: That's pretty much standard practice for auto-generated config files. See e.g. any `conf.py` in the sphinx docs.

kadi/tests/test_occweb.py
taldcroft: Done.

kadi/occweb.py
jeanconn: Right, it just seems like `get_occweb_page` could have the if statement and then not hand the username and password to a routine just to get them back.

docs/commands_v2.rst
```diff
@@ -0,0 +1,313 @@
...
+
+Custom scenario example
+^^^^^^^^^^^^^^^^^^^^^^^
+
+This example shows the steps to programmatically add an ACIS CTI in the midst
+of the 2021:296 NSM recovery::
+
+    # For Ska3 2022.2 this is necessary
+    >>> import sys
+    >>> sys.path.insert(0, '/Users/aldcroft/git/parse_cm')
```
jeanconn, line 97: I don't see how this PR is going to get promoted with the parse_cm one too so...

docs/commands_v2.rst
```diff
@@ -0,0 +1,313 @@
...
+    INFO: The configuration file has been successfully written to
+    ~/.kadi/config/kadi.cfg [astropy.config.configuration]
+
+
+Environment variables
+---------------------
+
+``KADI``
+  Override the default location of kadi flight data files ``cmds2.h5`` and
+  ``cmds.pkl``.
```
jeanconn, line 224: I assume this is still cmds2.pkl in this context, but that's a nit.

docs/commands_v2.rst
```diff
@@ -0,0 +1,313 @@
...
+
+``<scenario>/``
+  Directory containing files for a custom scenario. The files are
+  ``cmd_events.csv``, ``loads.csv``, ``loads.dat``. Note that the ``loads/``
+  directory is not specific to a scenario and so the top-level version is used.
+
+Web resources
+^^^^^^^^^^^^^
+
+`Chandra Command Events`_ she Google sheet
```
jeanconn, line 289: Typo "she" ?

kadi/commands/observations.py
```diff
@@ -0,0 +1,395 @@
...
+            raise ValueError(f'No matching observations for {obsid=}')
+
+    obss = [cmd['params'] for cmd in cmds_obs]
+    for obs, cmd_obs in zip(obss, cmds_obs):
+        obs['source'] = cmd_obs['source']
+
+    return obss
+
+
+def get_agasc_cone_fast(ra, dec, radius=1.5, date=None):
```
jeanconn, line 353: Should this just move into `agasc`?

kadi/paths.py
taldcroft: I really struggled with the name and probably have failed to convey the real meaning in the docs, but the `flight` config is really just a proxy for "ONLY use the current $SKA/data/kadi/cmds2.h5 and do not use network resources to get the latest updates to Command Events and/or approved loads". 

So running production load review on HEAD is one good use case, but another one is just to make things faster on a laptop by not bothering with the 5-10 seconds it can take to do the full update (downloading the Google sheet and hitting OCCweb approved products directories). I've done this myself already a few times.

For the most part I think checking for `/proj/sot/ska` is something we should avoid going forward. There may still be cases, but that isn't the best test of being on HEAD.

kadi/commands/command_sets.py
taldcroft: Fixed.

kadi/commands/command_sets.py
taldcroft: Yeah, this isn't the most beautiful. But adding a generic `kwargs` doesn't really help from what I can see. You still get the linter or reader of the code complaining that `kwargs` is not being used. As it stands it is clear that every `cmd_set_*` function must accept an optional `date` and nothing else. With `kwargs` then the function of that unused keyword is less clear.

kadi/commands/command_sets.py
taldcroft: Good catch, this is a leftover docstring from a previous version, fixed.

kadi/occweb.py
taldcroft: But this function handles a middle-ground case of `username` being `None`. So the API really needs to provide the user, passwd and return a user, passwd pair. 

It's really just simplest to have a single function that does *everything* instead of delegating some of that user/pass logic outside. And if we ever wrote another function that needs OCCweb auth handling then that function would need to repeat the special handling of both user and passwd being provided.

kadi/commands/observations.py
```diff
@@ -0,0 +1,395 @@
...
+OBSERVATIONS = {}
+
+# Cache of important columns in proseco_agasc_1p7.h5
+STARS_AGASC = None
+
+
+def set_detector_and_sim_offset(aca, simpos):
+    """Get detector from SIM position.
+
+    Finds the detector with nominal SIM position closest to ``simpos``.
```
jeanconn, line 45: From V&V I seem to recall exceptions to this (ACIS-I vs S) but I suppose it may not matter for this application.

kadi/commands/observations.py
```diff
@@ -0,0 +1,395 @@
...
+    stars = get_agasc_cone_fast(q_att.ra, q_att.dec, radius=1.2, date=aca.date)
+    yang_stars, zang_stars = radec_to_yagzag(stars['RA_PMCORR'], stars['DEC_PMCORR'], q_att)
+    idxs_aca = np.where(np.isin(aca['type'], ('ACQ', 'GUI', 'BOT')))[0]
+    for idx_aca in idxs_aca:
+        yang = aca['yang'][idx_aca]
+        zang = aca['zang'][idx_aca]
+        dys = np.abs(yang - yang_stars)
+        dzs = np.abs(zang - zang_stars)
+
+        # Get the brightest star within a box (default = 5 arcsec halfwidth)
```
jeanconn, line 111: Comment update if we're moving to 1.5.

taldcroft:
For review, the new commands `core.py` module started life as `kadi/commands/commands.py`, so you can see the diffs somewhat better with:
```
git show master:kadi/commands/commands.py > commands_master.py
opendiff commands-master.py kadi/commands/core.py
```

taldcroft:
@jeanconn - I added a more complete functional test of the star and fid identification in the top description.

kadi/commands/observations.py
taldcroft: Fixed.

kadi/commands/observations.py
taldcroft: Yes, doesn't matter here. Undercover observations will also be incorrectly labeled, but that's OK.

kadi/commands/observations.py
taldcroft: It's getting late, so unless this is explicitly wrong or bad I'd like to leave it. Maybe when we come across a 2nd application that needs this function we can refactor into the core agasc.

docs/commands_v2.rst
taldcroft: Fixed.

docs/commands_v2.rst
taldcroft: Fixed.

docs/commands_v2.rst
taldcroft: Fixed.

docs/commands_v2.rst
taldcroft: Fixed.

docs/commands_v2.rst
taldcroft: Fixed.

kadi/commands/commands_v2.py
```diff
@@ -0,0 +1,1185 @@
...
+    return cmds_obs
+
+
+# parameters from characteristics_Hardware for normal maneuvers
+MANVR_DELTA = 60.0
+MANVR_ALPHAMAX = 2.18166e-006  # AKA alpha
+MANVR_VMAX = 0.001309  # AKA omega
+
+
+def manvr_duration(q1, q2):
```
jeanconn, line 554: Similar to the agasc comment, this probably doesn't belong here.  But not a blocker to beta promotion.

kadi/commands/core.py
```diff
@@ -0,0 +1,824 @@
...
+            del params['type']
+
+    out = CommandTable(out)
+    out['time'].info.format = '.3f'
+
+    return out
+
+
+def _find(start=None, stop=None, inclusive_stop=False, idx_cmds=None,
+          pars_dict=None, **kwargs):
```
jeanconn, line 201: Might need some docstring help with idx_cmds and pars_dict.

kadi/commands/tests/test_commands.py
```diff
@@ -3,52 +3,99 @@
...
+    else:
+        assert np.all(cs['source'][:10] == 'JAN2612A')
+        assert np.all(cs['source'][-10:] == 'JAN3012C')
     assert cs['date'][0] == '2012:029:13:00:00.000'
     assert cs['date'][-1] == '2012:030:11:00:01.285'
     assert cs['tlmsid'][-1] == 'CTXBON'
 
-    cs = commands._find('2012:029:12:00:00', '2012:030:12:00:00', type='simtrans')
+    cs = core._find('2012:029:12:00:00', '2012:030:12:00:00', type='simtrans',
+                    idx_cmds=idx_cmds, pars_dict=pars_dict)
```
jeanconn, line 73: Might be safer to version these tests too (if version 1 etc) instead of using the idx_cmds and pars_dict.  Even if _find is not a public method.

docs/commands_states.rst
```diff
@@ -179,6 +192,288 @@ Notes and caveats
...
+   commands_v2.rst
+
+
+Getting commands
+^^^^^^^^^^^^^^^^
+
+In order to use the v2 version do the following::
+
+  >>> from kadi import commands
+  >>> commands.conf.commands_archive_version = "2"  # must be the string "2" not int 2
```
jeanconn, line 217: Is this "commands_archive_version" or "commands_version"?

docs/commands_states.rst
```diff
@@ -32,25 +44,26 @@ installed code.  Note that a key design feature is that is it straightforward fo
...
-  --------------------- ---------- ---------- --- ---- ----------- ------
-  2013:001:00:37:37.653   ORBPOINT       None   0    0   426098988    N/A
-  2013:001:00:53:07.181 COMMAND_SW   AOACRSTD 129 1524   426098990    N/A
-  2013:001:00:54:07.181 COMMAND_SW   AOFUNCDS 129 1526   426098990    N/A
-  2013:001:00:55:07.181 COMMAND_SW   AOFUNCDS 129 1528   426098990    N/A
-  2013:001:00:56:07.181 COMMAND_SW   AONMMODE 129 1530   426098990    N/A
-  2013:001:00:56:07.181    ACISPKT AA00000000 132 1620   426098991    N/A
-  2013:001:00:56:07.181   SIMTRANS       None 132 1623   426098991    N/A
-  2013:001:00:56:07.438 COMMAND_SW   AONM2NPE 129 1532   426098990    N/A
+          date            type      tlmsid   scs step      time     timeline_id   vcdu  params
```
jeanconn, line 56: It looks like the other examples previously didn't print with a "time" column as well, but that doesn't match the flight output.

jeanconn:
Should there be a test that hits the new stuff in parse_cm or are the unit tests over there sufficient?

docs/commands_v2.rst
```diff
@@ -0,0 +1,310 @@
...
+    PosixPath('/Users/aldcroft/.kadi/nsm-cti/cmd_events.csv')
+    >>> path_cti.parent.mkdir(exist_ok=True, parents=True)
+
+    >>> from astropy.table import Table
+    >>> events_flight = Table.read(path_flight)
+    >>> events_flight.colnames
+    ['State', 'Date', 'Event', 'Params', 'Author', 'Reviewer', 'Comment']
+
+    >>> cti_event = {'State': 'definitive', 'Date': '2021:297:13:00:00',
+    ...              'Event': 'RTS', 'Params': 'RTSLOAD,1_CTI06,NUM_HOURS=12:00:00,SCS_NUM=135',
```
jeanconn, line 117: I was a bit surprised that NUM_HOURS=4:00:00 wasn't compliant but not a biggie.

taldcroft: I can't find a spec, but having the leading zero is probably consistent with normal usage. In any case I updated the key FOT MP process guide TWiki page like so:

image

docs/commands_states.rst
taldcroft: Fixed.

docs/commands_states.rst
taldcroft: Yeah, I got a little lazy here and didn't update the rest of the `v1` examples outputs since `v1` will be deprecated and then removed.

kadi/commands/tests/test_commands.py
taldcroft: Sorry, I don't know what you mean, please try again.

kadi/commands/core.py
taldcroft: Done.

kadi/commands/__init__.py
taldcroft: As follow-up I did comprehensive functional testing for mission and 1.5 arcsec is fine.

kadi/commands/commands_v2.py
taldcroft: Copy.

taldcroft:
> Looking at the diff for core / commands, what were your thoughts on this bit

Great catch, that is an undesired API change for V1. I'll fix it. I also see how to fix `Chandra.Time.date2secs` so the original weirdness doesn't show up.

kadi/commands/tests/test_commands.py
jeanconn: What I was trying to say is that for many of the tests, the diff was easy to see as the "v1" version basically just copied the old test into an if/else.  Here, the code in the test test is now changed to pass idx_cmds and pars_dict (and what those are and what they do isn't really defined in _find()) so it is harder to see if the "v1" test is testing the "same" thing.

jeanconn:
Regarding "I'll also"... did you decide the fix in Chandra.Time was sufficient so we don't need to bring back the bytestring_to_unicode?

kadi/commands/tests/test_commands.py
taldcroft: OK I see. Because of the split V1 and V2, the new `idx_cmds` and `pars_dict` args are actually required args to `_find`. The values being provided in this test for V1 correspond precisely to what was hardwired in the old version (current release/master) of `_find` in the old `commands.py`. So it is not possible to replicate exactly in code the old test because of this API change, but examination of the old/new code shows easily that the test is doing the same thing.

jeanconn (approved):

PR #218, opened 2022-03-29
Fix regressions from v2 (#212)

taldcroft:
## Description

This fixes some regression (ska_testr) problems introduced by #212.

- Use absolute rather than relative imports. When I moved scripts into the `scripts` subpackage those relative imports were broken. Another reason to avoid relative imports in most (but not all) situations.
- Reverted the name of the command-line script from `kadi_update_cmds_v1` back to the original `kadi_update_cmds`. This should fix the error in ska_testr where it failed to find the expected script.
- Logging was doing unexpected things because of the addition of a new top-level logger with the name `kadi`. Modules that initialized a logger with `__name__` were generating output from two loggers. For some reason that is actually still a mystery to me, when I did a local install test of kadi the top-level logger was getting a log-level of DEBUG and generating unexpected logging output. I'd still like to track that down, but using new logger names that don't start with `kadi.` fixes this.

## Testing

- [x] Functional testing

### Functional testing

Installed into a test environment, copied the current (but a few days out of date) kadi data (cmds v1, v2 and events) into a  local test directory (not the git repo), and successfully ran:
```
kadi_update_cmds
kadi_update_cmds_v2
kadi_update_events
```

Then ran ska_testr in the test env and got a PASS (with https://github.com/sot/ska_testr/pull/51)
```
cd ~/git/ska_testr
run_testr --include kadi
```

### DIFF
--- a/kadi/cmds/cmds.py
+++ b/kadi/cmds/cmds.py
@@ -8,7 +8,7 @@ from Chandra.Time import DateTime
 import pickle
 import warnings
 
-from ..paths import IDX_CMDS_PATH, PARS_DICT_PATH
+from kadi.paths import IDX_CMDS_PATH, PARS_DICT_PATH
 
 __all__ = ['filter']
 
--- a/kadi/events/models.py
+++ b/kadi/events/models.py
@@ -2337,7 +2337,7 @@ class AsciiTableEvent(BaseEvent):
         Get events from telemetry defined by a simple rule that the value of
         `event_msids[0]` == `event_val`.
         """
-        from ..paths import DATA_DIR
+        from kadi.paths import DATA_DIR
 
         start = DateTime(start).date
         stop = DateTime(stop).date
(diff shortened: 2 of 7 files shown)


taldcroft:
My plan is to come back later with less time pressure and rework all the logging to do it consistently and correctly.

javierggt:
Is this PR supposed to fix these errors?
```
'warning' matched at test_regress.log:2217 :: /proj/sot/ska/jgonzalez/git/kadi/kadi/cmds/cmds.py:17: FutureWarning: kadi.cmds is deprecated and no longer tested, use kadi.commands instead
'warning' matched at test_regress.log:2218 :: warnings.warn('kadi.cmds is deprecated and no longer tested, use kadi.commands instead',
'error' matched at test_regress.log:2241 :: AttributeError: 'LazyVal' object has no attribute '_val'
'error' matched at test_regress.log:2266 :: raise IOError("``%s`` does not exist" % (filename,))
'error' matched at test_regress.log:2267 :: OSError: ``/export/jgonzale/ska_testr/outputs/logs/Linux_2022-03-29T18-39-49_8baabe2_kady.cfa.harvard.edu/kadi/cmds.h5`` does not exist
```

PR #221, opened 2022-04-07
Add starcat_date to observation parameters + fix caching and minor issues

taldcroft:
## Description

Add `starcat_date` to the parameters stored in observations.

Other things along the way:
- Noticed that the `clear_caches` was missing the OBSERVATIONS cache, so that was fixed.
- Noticed that the result from `get_observations` was a reference to the original params dicts in cache, so this allowed for shenanigans if the user played with those outputs. This, of course, came up in testing and took quite a while to figure out.
- Added some `clear_cache()` commands in certain tests to ensure independence of results from test order.
- Updated to run most of the `get_observations` or `get_starcats` test using the `flight` scenario so they will run on GRETA.

Also includes #225 and #226.

Fixes #220

## Interface impacts

- Adds a new key `starcat_date` to the dict returned by `get_observations`
- Changes the star catalog `date` from being the `obs_start` (start of NPNT) to `starcat_date` (commanded star catalog time).
- Requires new versions of `cmds2.pkl` and `cmds2.h5`. I've generated these and will place them on HEAD and GRETA. The new data files are expected to be compatible with the release 5.9.2 version in 2022.3 but I'll check that.

Running the 5.9.2 unit tests with the new data files gives 3 failures, all expected:
```
=================================================== short test summary info ====================================================
FAILED kadi/commands/tests/test_commands.py::test_commands_create_archive_regress[2] - AssertionError: assert False
FAILED kadi/commands/tests/test_commands.py::test_get_observations_by_obsid_single - AssertionError: assert [{'manvr_star...2...
FAILED kadi/commands/tests/test_commands.py::test_get_observations_by_obsid_multi - AssertionError: assert [{'manvr_star...00...
======================================== 3 failed, 50 passed, 103 deselected in 31.06s =========================================
```

### Unit tests

- [x] Mac

Independent check of unit tests by Jean
- [x] Linux

### Functional tests

Jean functionally confirmed the `unique` option to get_starcats_as_table with this query that gets the two catalogs (same catalog) of an ACIS undercover series.
```
In [5]: get_starcats_as_table(start='2020:366:17:04:20.000', stop='2020:366:23:00:00.000')
Out[5]: 
<Table length=22>
obsid      starcat_date      slot  idx      id    type  sz         mag          maxmag          yang                zang         dim   res  halfw
int64         str21         int64 int64   int64   str3 str3      float64       float64        float64             float64       int64 int64 int64
[...]

PR #222, opened 2022-04-08
Fix bug in command-not-run command set and test

taldcroft:
## Description

There was a problem with the `Command not run` command set that was missed in testing. It had previously worked when the `cmds2.h5` was generated, but a subsequent change broke it, but this was not seen until I regenerated the commands now.

This PR fixes the problem and adds complete testing of all the allowed command sets.

Note: in order for tests to pass this is based off of #221 and the merge request is into that branch.

## Interface impacts

None.

### Unit tests

- [x] Mac

Independent check of unit tests by Javier
- [x] Mac

### Functional tests

No functional testing.

### DIFF
--- a/kadi/commands/command_sets.py
+++ b/kadi/commands/command_sets.py
@@ -194,7 +194,7 @@ def get_cmds_from_event(date, event, params_str):
     if isinstance(params_str, str):
         if event == 'RTS':
             args = [params_str]
-        elif event == 'Command':
+        elif event in ('Command', 'Command not run'):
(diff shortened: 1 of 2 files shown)


javierggt (approved):
The changes are simple enough, and I verified that the test introduced in this PR (`kadi/commands/tests/test_commands.py::test_get_cmds_from_event_all`) fails with the previous version of `command_sets.py`, so this PR indeed fixes the bug being tested for.

All 167 unit tests also passed

PR #231, opened 2022-04-27
Web-free kadi

taldcroft:
## Description

This removes the web server part of the kadi package, instead making kadi events be strictly a standalone django ORM package.

Importantly, this removes the `django.setup()` from the `kadi.events` init and puts it into `query.py`. At the same time the symbols from `query.py` are only lazily imported on demand. This allows two things to work:
- Making (migrating) the ORM database via `manage.py`, which itself calls `django.setup()` and subsequently imports `kadi.events.models`. Previously the code used a non-public interface to avoid a second call to `django.setup()`.
- Any user query e.g. `kadi.events.manvrs.filter(...)` will autoload `query.py` the first such query and run `django.setup()` at that point.

The one downside of the setup reorganization is that the list of available event queries needs to be hard-coded in `kadi/events/__init__.py`. I could not find a way around this, but given the stability of the event types this is not a problem in practice.

To do:
- [x] Test that this works in the new flask web-kadi. Does it fix the 500 errors?? [YES, see below]
- [x] A bit more functional testing

## Interface impacts

None. This still needs testing but that is the intent.

### Unit tests

- [x] Mac (using the current flight database)

- [ ] [PLATFORM]:

#### Test `update_events.py` script

Rebuilt the `events3.db3` database following the steps in `NOTES.build`.
```
cd ~/git/kadi
export KADI=$PWD
rm -f events3.db3
rm -rf kadi/events/migrations
./manage.py makemigrations events
./manage.py migrate

python -m kadi.scripts.update_events --start=1999:001 --stop=2000:001
python -m kadi.scripts.update_events --start=2000:001
```

- [ ] Run unit tests with `KADI` still pointing to the local file, PASS.

#### Module lazy imports
```
In [1]: from kadi import events

In [2]: dir(events)
Out[2]: 
['caps',
 'dark_cal_replicas',
 'dark_cals',
 'dsn_comms',
 'dumps',
 'dwells',
 'eclipses',
 'fa_moves',
 'grating_moves',
 'load_segments',
 'ltt_bads',
 'major_events',
 'manvrs',
 'models',
 'normal_suns',
 'obsids',
 'orbits',
 'pass_plans',
 'rad_zones',
 'safe_suns',
 'scs107s',
 'tsc_moves']

In [3]: events.<TAB>  # Works as expected and shows the above list

In [4]: from kadi.events import *

In [5]: whos
Variable            Type                Data/Info
-------------------------------------------------
caps                EventQuery          <EventQuery: CAP>
dark_cal_replicas   EventQuery          <EventQuery: DarkCalReplica>
dark_cals           EventQuery          <EventQuery: DarkCal>
dsn_comms           EventQuery          <EventQuery: DsnComm>
dumps               EventQuery          <EventQuery: Dump>
dwells              EventQuery          <EventQuery: Dwell>
eclipses            EventQuery          <EventQuery: Eclipse>
events              module              <module 'kadi.events' fro<...>kadi/events/__init__.py'>
fa_moves            EventQuery          <EventQuery: FaMove>
grating_moves       EventQuery          <EventQuery: GratingMove>
load_segments       EventQuery          <EventQuery: LoadSegment>
ltt_bads            LttBadEventQuery    <EventQuery: LttBad>
major_events        EventQuery          <EventQuery: MajorEvent>
manvrs              EventQuery          <EventQuery: Manvr>
models              module              <module 'kadi.events.mode<...>i/kadi/events/models.py'>
normal_suns         EventQuery          <EventQuery: NormalSun>
obsids              EventQuery          <EventQuery: Obsid>
orbits              EventQuery          <EventQuery: Orbit>
pass_plans          EventQuery          <EventQuery: PassPlan>
rad_zones           EventQuery          <EventQuery: RadZone>
safe_suns           EventQuery          <EventQuery: SafeSun>
scs107s             EventQuery          <EventQuery: Scs107>
tsc_moves           EventQuery          <EventQuery: TscMove>
```


taldcroft:
@javierggt - can you give this a spin in web-kadi-test and see if it fixes the intermittent 500 errors?

kadi/events/query.py
```diff
@@ -6,6 +6,9 @@
 
 from Chandra.Time import DateTime
 
+import django
+django.setup()
+
```
javierggt, line 11: This is the magic that makes django work.

This being here causes an error if one imports kadi.events.models befare kadi.events.query:
```python
In [1]: from kadi.events import models
---------------------------------------------------------------------------
AppRegistryNotReady                       Traceback (most recent call last)
<ipython-input-1-0caab80e9759> in <module>
----> 1 from kadi.events import models

~/git/kadi/kadi/events/models.py in <module>
    239 
    240 
--> 241 class Update(models.Model):
    242     """
    243     Last telemetry which was searched for an update.

/export/servers/kadi-test/ska3/lib/python3.8/site-packages/django/db/models/base.py in __new__(cls, name, bases, attrs, **kwargs)
    106 
    107         # Look for an application configuration to attach the model to.
--> 108         app_config = apps.get_containing_app_config(module)
    109 
    110         if getattr(meta, 'app_label', None) is None:

/export/servers/kadi-test/ska3/lib/python3.8/site-packages/django/apps/registry.py in get_containing_app_config(self, object_name)
    251         Return None if the object isn't in any registered app config.
    252         """
--> 253         self.check_apps_ready()
    254         candidates = []
    255         for app_config in self.app_configs.values():

/export/servers/kadi-test/ska3/lib/python3.8/site-packages/django/apps/registry.py in check_apps_ready(self)
    134             # exception.
    135             settings.INSTALLED_APPS
--> 136             raise AppRegistryNotReady("Apps aren't loaded yet.")
    137 
    138     def check_models_ready(self):

AppRegistryNotReady: Apps aren't loaded yet.

In [2]: from kadi.events import query

In [3]: from kadi.events import models
```

javierggt:
This branch is now installed in the web-kadi-test environment, so it is running on kadi-test. The usual error has not shown up yet. We should just start using it more regularly to see.

kadi/events/query.py
taldcroft: Good catch. It's like whack-a-mole. I'm just getting very annoyed with django, but the whole kadi events are pretty deeply integrated into that framework.

At some level users don't need to import `kadi.events.model` (or could import query first) but I agree it has a bad taste.

taldcroft: Well I was hitting the server for a while with no errors and feeling good, until:

{"error": "Traceback (most recent call last):\n  File \"/export/servers/config/test/wsgi-scripts/api.py\", line 66, in application\n    app_func, app_kwargs, server_opt = parse_url(environ)\n  File \"/export/servers/config/test/wsgi-scripts/api.py\", line 142, in parse_url\n    app_func = getattr(app_func, func_part)\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/site-packages/kadi/events/__init__.py\", line 84, in __getattr__\n    query = importlib.import_module('kadi.events.query')\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/importlib/__init__.py\", line 127, in import_module\n    return _bootstrap._gcd_import(name[level:], package, level)\n  File \"<frozen importlib._bootstrap>\", line 1014, in _gcd_import\n  File \"<frozen importlib._bootstrap>\", line 991, in _find_and_load\n  File \"<frozen importlib._bootstrap>\", line 975, in _find_and_load_unlocked\n  File \"<frozen importlib._bootstrap>\", line 671, in _load_unlocked\n  File \"<frozen importlib._bootstrap_external>\", line 843, in exec_module\n  File \"<frozen importlib._bootstrap>\", line 219, in _call_with_frames_removed\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/site-packages/kadi/events/query.py\", line 10, in <module>\n    django.setup()\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/site-packages/django/__init__.py\", line 24, in setup\n    apps.populate(settings.INSTALLED_APPS)\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/site-packages/django/apps/registry.py\", line 114, in populate\n    app_config.import_models()\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/site-packages/django/apps/config.py\", line 211, in import_models\n    self.models_module = import_module(models_module_name)\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/importlib/__init__.py\", line 127, in import_module\n    return _bootstrap._gcd_import(name[level:], package, level)\n  File \"<frozen importlib._bootstrap>\", line 1014, in _gcd_import\n  File \"<frozen importlib._bootstrap>\", line 991, in _find_and_load\n  File \"<frozen importlib._bootstrap>\", line 975, in _find_and_load_unlocked\n  File \"<frozen importlib._bootstrap>\", line 671, in _load_unlocked\n  File \"<frozen importlib._bootstrap_external>\", line 843, in exec_module\n  File \"<frozen importlib._bootstrap>\", line 219, in _call_with_frames_removed\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/site-packages/kadi/events/models.py\", line 9, in <module>\n    django.setup()\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/site-packages/django/__init__.py\", line 24, in setup\n    apps.populate(settings.INSTALLED_APPS)\n  File \"/export/servers/kadi-test/ska3/lib/python3.8/site-packages/django/apps/registry.py\", line 83, in populate\n    raise RuntimeError(\"populate() isn't reentrant\")\nRuntimeError: populate() isn't reentrant\n"}

javierggt: I guess you either already know this or are googling it right now...

https://blog.whtsky.me/tech/2020/populate_isnt_reentrant

javierggt: In my local copy I just added the call to `django.setup()` inside `kadi.models`. I don't know why it worked fine for a while.

taldcroft: Yes, I've been there, done that.. :smile:  Probably more useful is the first answer in this SO: https://stackoverflow.com/questions/27093746/django-stops-working-with-runtimeerror-populate-isnt-reentrant.

The problem is that the issue only shows up on the Apache web server. So maybe we could monkeypatch the web server django to get a better traceback?

About adding `django.setup()` inside `kadi.models`, this will break `manage.py` because that calls `setup()` and then imports `kadi.models` and you reliably get the reentrant exception. `manage.py` is still required in order to build or migrate the database models.

In any case, my best effort of having only one place in the kadi code that does `django.setup()` has failed to stop the reentrant problem. One guess is that this is related to Apache running Python sub-interpreters so that something really funky is happening.

My only answer now is to just punt on the kadi events web-kadi API service. AFAIK there are zero actual users. We test it a lot but I don't think it is used for actual work.

javierggt: as to where to put the call to `django.setup()`, you can have it inside a little module that only does that and is imported from all others, essentially letting python do the lazy load, depending on which module is imported first. That would not prevent funky business with WSGI sub-interpreters but would fix our usual use cases (importing `query`, `models` or running `manage.py`.

This would still be the point where we would have crazy logic to try to prevent double loading and such.

taldcroft: @javierggt - I pushed a new commit to catch the reentrant exception. Can you try installing that to web-kadi-test?

javierggt: I removed the change I made to add the call to setup inside `models`. That might actually have been the cause of the error you found.

There are two tests I can make before monkeypatching django in kadi-test:
- make sure the top-level app imports `kadi.event.query`
- set the process group to `%{GLOBAL}`

taldcroft:
I just started a script to just pound on the kadi events API. After 5 minutes it is still going with no errors.
```
import sys
import time
import numpy as np
import requests
from cxotime import CxoTime

queries = [
    'obsids',
    'tsc_moves',
    'dark_cal_replicas',
    'dark_cals',
    'scs107s',
    'fa_moves',
    'grating_moves',
    'dumps',
    'eclipses',
    'manvrs',
    'dwells',
    'safe_suns',
    'normal_suns',
    'major_events',
    'caps',
    'load_segments',
    'pass_plans',
    'dsn_comms',
    'orbits',
    'rad_zones',
    'ltt_bads'
]

n_req = 0
while True:
    start = CxoTime('2021:001', out_subfmt='date') + np.random.randint(0, 300)
    stop = start + 3
    stop.out_subfmt = 'date'
    query = np.random.choice(queries)
    url = f'https://web-kadi-test.cfa.harvard.edu/api/kadi/events/{query}/filter?start={start}&stop={stop}'
    t0 = time.time()
    rq = requests.get(url)
    n_req += 1
    t1 = time.time()
    if rq.status_code != 200:
        print()
        print('*' * 30, 'Bad request', '*' * 30)
        print(rq.status_code)
        print(rq.text)

    try:
        rq.json()
        print(query[0], end='')
        sys.stdout.flush()
    except Exception as exc:
        print()
        print('*' * 30, 'Bad JSON', '*' * 30)
        print(exc)
```

javierggt:
I think that looks good. Most probably the error was because the way I made my local change.

We could catch the `AppRegistryNotReady` at the top of `kadi.events.models` and issue a more informative error.

I checked, and the following seem to import `kadi.events.models`:
- `kadi/scripts/update_events.py`

taldcroft:
> I checked, and the following seem to import kadi.events.models

Great catch. I had at one point tested that the `update_events` script works, but it was indeed broken with an `AppRegistryNotReady` exception. But (finally something that works), I discovered a minimal workaround that seems to do the trick. See 1fba5d9 and the very long comment there.

I'm running the steps in `NOTES.build` now to rebuild the entire db from scratch and so far all looks good.

kadi/events/__init__.py
```diff
@@ -42,22 +42,66 @@
...
+# There is a little subtlety here. See discussion in
+# https://github.com/sot/kadi/pull/231, but the upshot is that we need
+# `from kadi.events import models` to run `django.setup()`. This syntax is what
+# gets used in `update_events.py`. By including 'models' in `__all__` that
+# imports it via `query.py` and thus runs django setup.
+#
+# `manage.py` imports the same module but runs `django.setup()` before importing
+# it. However, it appears to import this via the equivalent of
+# `import kadi.events.models`, which does NOT go through `query.py`.
+# With this trick of lazy loading, both pathways work!
```
javierggt, line 66: but the upshot is that `django.setup()` must be called only once. Within kadi, `django.setup()` is called in `kadi.query`. Where the call to `django.setup()` actually happens depends on the application:
- `manage.py` calls `django.setup()` and then imports kadi.models (and it never imports `kadi.query`)
- `update_events.py` does `from kadi.events import models`, which calls `events. __getattr__`, which in turn imports `kadi.query`
- in any case, `kadi.query` should be imported before models can be used.

PR #233, opened 2022-05-12
Fix lucky testing and occweb lucky functionality

taldcroft:
## Description

As part of FST-requested testing of the new lucky server, I noticed that testing of the `kadi.occweb` module was inadvertently skipping a test. The tests were written assuming HEAD linux where `lucky` resolves to the right FTP server, but this is not the case for a Mac laptop. Instead this provides the fully-resolved host name.

## Interface impacts

None

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

No functional testing.


javierggt (approved):
The changes seem pretty straight forward to me.

I did the following verification:
- on the current kadi release, run this test and verify it is skipped
   ```
   pytest kadi/tests/test_occweb.py::test_put_get_user_none
   ```
- I had to disconnect from the VPN for that.
- switch to the `fix-lucky` branch
- run the test again and verify it runs and passes

PR #240, opened 2022-06-08
Better exception message in get_states for too-early start arg

### LINKED ISSUE #239: states.get_states fails if you don't know the time of the start of the data
https://sot.github.io/kadi/commands_states.html  says:

"The Commands archive is a table of every load command that has been run, or is currently approved to be run, on the spacecraft since 2002."

If I want to capture data from the very beginning of the mission, I cannot just write:

In [63]: states.get_states('2002:001', '2003:001', state_keys = ['orbit_point'])

You get a value error:

**ValueError: did not find transitions for state key(s) {'orbit_point'} within 1000 days of 2002:001:00:00:00.000.  Maybe adjust the `lookbacks` argument?**

Firstly, I cannot adjust the "lookbacks" argument for this command:

In [64]: states.get_states('2002:001', '2003:001', state_keys = ['orbit_point'], lookbacks = (7, 15))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-64-c0815e040e22> in <module>
----> 1 states.get_states('2002:001', '2003:001', state_keys = ['orbit_point'], lookbacks = (7, 15))

**TypeError: get_states() got an unexpected keyword argument 'lookbacks'**

So that error message is misleading.

By experimentation I got close to the beginning:

states.get_states('2002:007:17:40:00.20', '2003:025', state_keys = ['orbit_point'])

Which will be good enough for my purposes but I shouldn't have to do that.

It should run like the SKA engineering archive wherein if you put a start time prior to the earliest data, it will give you the earliest data available.

If there is already a way to do this please let me know.

taldcroft:
## Description

This improves the exception message to provide better feedback to the user in the case when an early `start` time makes it not possible to establish continuity.

The suggestion in #239 was to make it that the returned states start when available data allow, similar to how `cheta` functions when fetching telemetry. This is a good idea, but due to the way dynamic command states are built up on the fly from continuity, implementing this would not be straightforward. The effective start time depends on when all of the required commands to define a particular state or states have been issued.

Instead, this just gets the user on the right track more quickly by highlighting the problem and a workaround.

Fixes #239

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

```
In [1]: from kadi.commands import states

In [2]: states.get_states('2002:001', '2003:001', state_keys = ['orbit_point'])
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/git/kadi/kadi/commands/states.py in get_states(start, stop, state_keys, cmds, continuity, reduce, merge_identical, scenario)
   1331         try:
-> 1332             continuity = get_continuity(start, state_keys, scenario=scenario)
   1333         except ValueError as exc:

~/git/kadi/kadi/commands/states.py in get_continuity(date, state_keys, lookbacks, scenario)
   1608         if missing_keys:
-> 1609             raise ValueError('did not find transitions for state key(s)'
   1610                              ' {} within {} days of {}.  Maybe adjust the `lookbacks` argument?'

ValueError: did not find transitions for state key(s) {'orbit_point'} within 1000 days of 2002:001:00:00:00.000.  Maybe adjust the `lookbacks` argument?

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-2-bccec27564a3> in <module>
----> 1 states.get_states('2002:001', '2003:001', state_keys = ['orbit_point'])

~/git/kadi/kadi/commands/states.py in get_states(start, stop, state_keys, cmds, continuity, reduce, merge_identical, scenario)
   1333         except ValueError as exc:
   1334             if 'did not find transitions' in str(exc):
-> 1335                 raise ValueError(
   1336                     f'no continuity found for {start=}. Need to have state '
   1337                     f'transitions following first command at {cmds[0]["date"]} '

ValueError: no continuity found for start='2002:001:00:00:00.000'. Need to have state transitions following first command at 2002:007:13:35:57.743 so use a later start date.
```

### DIFF
--- a/kadi/commands/states.py
+++ b/kadi/commands/states.py
@@ -1328,7 +1328,17 @@ def get_states(start=None, stop=None, state_keys=None, cmds=None, continuity=Non
 
     # Get initial state at start of commands
     if continuity is None:
-        continuity = get_continuity(start, state_keys, scenario=scenario)
+        try:
+            continuity = get_continuity(start, state_keys, scenario=scenario)
+        except ValueError as exc:
+            if 'did not find transitions' in str(exc):
+                raise ValueError(
+                    f'no continuity found for {start=}. Need to have state '
+                    f'transitions following first command at {cmds[0]["date"]} '
+                    'so use a later start date.'
+                )
+            else:
+                raise
 
     # Get transitions, which is a list of dict (state key
     # and new state value at that date).  This goes through each active
--- a/kadi/commands/tests/test_states.py
+++ b/kadi/commands/tests/test_states.py
@@ -1507,3 +1507,8 @@ def test_grating_motion_states():
            '2021:230:00:37:56.002 2021:230:00:41:19.002 RETR_MOVE      RETR    NONE grating,letg',
            '2021:230:00:41:19.002 2021:230:12:00:00.000      RETR      RETR    NONE         letg']
     assert sts.pformat_all() == exp
+
+
+def test_early_start_exception():
+    with pytest.raises(ValueError, match="no continuity found for start='2002:001:00:00:00.000'"):
+        states.get_states('2002:001', '2003:001', state_keys=['orbit_point'])


Gregg140:
Why not also figure out the earliest time one can search for states and put that in the documentation as well.  Right now the documentation says:
-----------------------------------------------------------------
Commands and states

Commands

The Commands archive is a table of every load command that has been run, or is currently approved to be run, on the spacecraft since **2002**.
--------------------------------------------------------------------

Figure out the earliest time I can put for start date in a  get_states call and put it right there at the top of the docs
instead of 2022.

I doubt it will change.

And also put it in the error messages.

Gregg140 (approved):
The earliest possible date a user can specify, where all states have been populated, ought to be listed in the documentation.

taldcroft:
> The earliest possible date a user can specify, where all states have been populated, ought to be listed in the documentation.

The problem is that "all states" is not well-defined. Since states are dynamically computed and there are many of them (including more obscure ones that are not in the legacy standard list), there is no single answer if you want to push the boundaries.

I could conservatively say that a query starting in 2003:001 will always succeed and declare that as the beginning.

PR #243, opened 2022-07-08
Add EventQuery to symbols exported from query.py by events/__init__.py

taldcroft:
## Description

`ska_testr` integration showed a script that was expecting `EventQuery` to be available as an import from `kadi.events`, so this adds that in.

## Interface impacts

Fixes an unexpected API change where `EventQuery` was not available.

### Unit tests

- [x] Mac

### Functional tests

With `PYTHONPATH` set to the git repo at this branch, I ran `run_testr --include kadi`:
- Without the fix (current master) I reproduced the failure related to `EventQuery` seen in ska3-flight 2022.7rc2.
- With the fix all tests pass.

### DIFF
--- a/kadi/events/__init__.py
+++ b/kadi/events/__init__.py
@@ -68,6 +68,7 @@ os.environ["DJANGO_ALLOW_ASYNC_UNSAFE"] = "true"
 # With this trick of lazy loading, both pathways work!
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


javierggt (approved):
I ran testr and varified the tests pass with this PR and fail with master.

PR #247, opened 2022-08-23
Fix bug where RLTT from not-run load is still applied

taldcroft:
## Description

This fixes a bug in commands v2 where the RLTT from a not-run load (marked as `Load not run` in the Command Events sheet) was still being applied to other commands.

In addition:
- Fix a problem where a set of commands that were filtered by the RLTT could cause an index exception.
- Provide info about the RLTT for each load in logging.

This fix impacted an existing test where the expected answer was actually wrong since part of the ACIS ECS commanding had been cut by the RLTT of OCT2521A which was never run.

## Interface impacts

None

### Unit tests

- [x] Mac (with new unit test)

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/kadi/commands/commands_v2.py
+++ b/kadi/commands/commands_v2.py
@@ -328,7 +328,10 @@ def update_archive_and_get_cmds_recent(
         # Command Events sheet).
         cmds = interrupt_load_commands(load, cmds)
         if len(cmds) > 0:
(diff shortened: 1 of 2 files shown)


taldcroft:
Superceded by #248

PR #248, opened 2022-08-24
Major refactor/improvement in command interrupt handling

taldcroft:
## Description

This implements an idea that has been around for a long time to handle command interrupt processing in a more direct -way that more closely matches the OBC behavior. In particular:
- Notice END SCS commands (CODISASX) and cut all running commands in that SCS slot after the command time.
- Implement RLTT as END SCS for slots 128 to 133.
- Put appropriate END SCS commands in the Command Event definitions.

A key implementation detail is the definition of "running commands". The basic outline of how this works is as follows.
- Make a list of `CommandTable` objects corresponding to each approved command load
- Append to that list a `CommandTable` of commands for each Command Event.
- Sort the list of `CommandTable`s (called `cmds_list`) by the date of the first command in each table.
- Iterate over `cmds_list` (`for cmds in cmds_list:`)
  - Find any RLTT or END SCS commands in `cmds` and collect them into a dict that captures the SCS numbers and end dates.
  - Iterate over all previous `cmds` in the `cmds_list` and cut commands accordingly.
- Stack the `cmds_list` tables and then sort into the standard backstop order.

The key point to this change is that it simplifies the code logic and removes ad-hoc processing rules.

To do:

- [x] Confirm that `loads.csv` is no longer used as an input to processing.
- [x] Consider further cleanup in code and maybe remove the `loads.csv` and `loads.dat` output. 
- [x] Check command history near 2022:223 BSH. Dither stays enabled?

## Interface impacts

New END SCS and LOAD_EVENT commands are stored in the archive. No changes to actual spacecraft commands.

### Unit tests

- [x] Mac with current "flight" archive (`$SKA/data/kadi/cmds2.*`)
- [x] Mac with new archive created using this code.

- [x] Linux

### Functional tests

- [x] Regenerate full-mission commands archive and confirm only expected diffs. See [validate/validate-cmd-interrupt-refactor.ipynb](https://github.com/sot/kadi/pull/248/files#diff-3f20ebeed41e7e1da2cee96d5f0f0bfc5bebfa2207b381af23cf8d7a45a65a8d)


kadi/commands/commands_v2.py
```diff
@@ -353,43 +359,61 @@ def update_archive_and_get_cmds_recent(
...
+    # by stopping SCS's 128-133.
+    for ii, cmds in enumerate(cmds_list):
+        cmds_source = cmds["source"][0]
+        if cmds_source == "CMD_EVT":
+            cmds_source = f"CMD_EVT {cmds[0]['event']} at {cmds[0]['date']}"
+        logger.info(f"Processing {cmds_source} with {len(cmds)} commands")
+        end_scs = collections.defaultdict(list)
+        if date_end := cmds.meta.get("rltt"):
+            source = f'RLTT in {cmds["source"][0]}'
+            end_scs[date_end, source].extend([128, 129, 130, 131, 132, 133])
```
jeanconn, line 382: Makes sense to me to interpret RLTT as a "stop" for those SCSs, but is that the definition?  I've forgotten and don't know where to find that definition.

kadi/commands/commands_v2.py
```diff
@@ -353,43 +359,61 @@ def update_archive_and_get_cmds_recent(
...
+
+        # Explicit END SCS commands. Most commonly these come from command events
+        # like SCS-107, but can also be in weekly loads e.g. disabling an ACIS
+        # ECS measurement in SCS-135.
+        ok = cmds["tlmsid"] == "CODISASX"
+        if np.any(ok):
+            for cmd in cmds[ok]:
+                if (scs := cmd["params"]["codisas1"]) >= 128:
```
jeanconn, line 390: Why are these limited to >= 128 here?  Because that's where they start or none below that would be relevant?

kadi/commands/commands_v2.py
taldcroft: Yes that's the definition. Running Loads Termination Time, where running loads are the weekly loads in SCSs 128-133 inclusive.

kadi/commands/commands_v2.py
taldcroft: None below 128 would be relevant. The SCS slots below 128 are protected SCS's that are in the OBC protected memory and require an OBC patch to change. The commands archive does not deal with starting and stopping those.

jeanconn: OK, so is this in here to be more self-documenting?  Because there also aren't any of those < 128 commands in here to be filtered, right?

taldcroft: There are SCS 98 disables in the weekly loads.

jeanconn: Right, but no scs=98 commands _in_ the loads.

kadi/commands/tests/test_commands.py
```diff
@@ -502,6 +531,24 @@ def test_cmds_scenario(stop_date_2020_12_03):
 def test_command_set_bsh():
     cmds = get_cmds_from_event("2000:001", "Bright star hold", "")
     exp = [
+        "2000:001:00:00:00.000 | COMMAND_SW       | CODISASX   | CMD_EVT  | "
+        "event=Bright_star_hold, event_date=2000:001:00:00:00, msid=CODISASX, "
+        "codisas1=12",
```
jeanconn, line 536: These look a little odd (and don't fully confirm the right behavior) due to truncation, though that's more of a problem with pformat_like_backstop than this PR.

jeanconn: I think technically this test should just be run using max_params_width=None for pformat_like_backstop.

kadi/commands/commands_v2.py
taldcroft: Correct.

kadi/commands/tests/test_commands.py
taldcroft: Fixed.

jeanconn (approved):

PR #255, opened 2022-10-03
Suppress v1 FutureWarning for testing

taldcroft:
## Description

For testing it is still allowed to import and use commands v1.

## Interface impacts

None

### Unit tests

- [ ] No unit tests
- [x] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

Using this branch with v7.0 version of the cmds2.* files passes `pytest` unit tests and does not issue any warning messages at the end.

If I change the `message=...` to something that does not match then I confirm the warnings are flagged by pytest.

### DIFF
--- a/kadi/commands/tests/test_commands.py
+++ b/kadi/commands/tests/test_commands.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 from pathlib import Path
 
 import astropy.units as u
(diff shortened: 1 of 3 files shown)


javierggt (approved):
I wouldn't have removed the line from `pytest.ini`, but that is a detail.

taldcroft:
> I wouldn't have removed the line from pytest.ini, but that is a detail.

I don't think we should suppress warnings that we do not expect. Getting that warning would indicate something wrong, so best to know about it from command-line `pytest`.

PR #259, opened 2022-10-28
Make noodle path be case insensitive

taldcroft:
## Description

Make life a little easier if a noodle path has the wrong case.

### Unit tests

- [x] Mac

- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/kadi/occweb.py
+++ b/kadi/occweb.py
@@ -253,8 +253,11 @@ def get_occweb_page(
         path = path.replace("\\", "/")
         if path.startswith("//noodle/"):
             for noodle_prefix, occweb_prefix in NOODLE_OCCWEB_MAP.items():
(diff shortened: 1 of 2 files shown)


jeanconn:
I thought this might make the whole path case-insensitive, but it looks like this just means that these
```
      \\noodle\FOT
      \\noodle\GRETA\mission\Backstop
      \\noodle\vweb
```
would be case-insensitive as the rest of the path needs case.  That's fine.  Since the case-insensitive part isn't described as a new feature in the help it seems fine this is just a bonus for a few paths (or parts of the path).

taldcroft:
@jeanconn - yes, this is a pretty limited fix. So please hit the Approve button if you are good with it.

jeanconn (approved):

PR #262, opened 2022-11-30
Remove WARNING in logging text

taldcroft:
## Description

This removes the word `WARNING` from the logging text for a situation that is common in anomaly recovery scenarios. This will reduce unnecessary alerts in Ska job watch.

## Interface impacts

None

## Testing

No testing given the scope of the change.


jeanconn:
For anomaly situations lately this has been obsid=0 or -1.  Should those be special cased and the rest still be warnings?

taldcroft:
That's only for anomalies that reset the OBC. If we have a BSH we'll still have this situation showing a processing error every day for a month.

jeanconn (approved):

PR #263, opened 2022-12-05
Explicit complete_intervals=True in logical intervals + code tidy

taldcroft:
## Description

This adds `complete_intervals=True` to a key call to `logical_intervals` that defines the creation of kadi events. Previously this was relying on the default value of that kwarg being True. Since that default is being changed in https://github.com/sot/eng_archive/pull/243, here we need to be explicit.

While in the code I noticed that an ugly hack of the `@import_ska` decorator is no long needed now that the kadi package does not provide a web server. This hack doesn't play well with IDEs (code introspection doesn't work).

## Interface impacts

None.

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/kadi/events/models.py
+++ b/kadi/events/models.py
@@ -704,7 +704,9 @@ class TlmEvent(Event):
             # Telemetry values for event_msids[0] define the states.  Don't allow a logical
             # interval that spans a telemetry gap of more than 10 major frames.
             times, bools = cls.get_state_times_bools(event_msidset)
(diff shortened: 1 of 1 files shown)


javierggt (approved):
The changes seem good to me. At first I looked at the diffs "color-blindly" (not noticing the `import_ska` part was red) and started wondering why you were adding that. Was happy to realize you were *removing* it.

I ran the unit tests, but I would note that the unit tests would not have caught the change in eng_archive.

PR #265, opened 2022-12-18
Fix issue with cmd query around 30 days before current time

taldcroft:
## Description

This fixes a bug in the `get_cmds` query logic which resulted in missing commands from the query under somewhat unusual circumstances:
- Query start is between 30 to 44 days before current time
- Within that window there is a non-load command from a Command Event.
- Query start is later than that non-load command time + 3 days.

The basic problem is that the code was checking for the first command of any type in the recent commands. Instead it should have been checking for the first load command in recent commands (within 30 days).

The unit test demonstrates the problem as it applies for a current date of 2022:352. In this case the OORMPEN non-load command at 2022:312:03:09:55.000 causes the problem.

## Interface impacts

None

## Testing

Added unit test which failed prior to the fix and passes now.

### Unit tests

- [x] Mac

- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/kadi/commands/commands_v2.py
+++ b/kadi/commands/commands_v2.py
@@ -225,7 +225,7 @@ def get_cmds(
             # Query does not overlap with recent commands, just use archive.
             logger.info("Getting commands from archive only")
             cmds = IDX_CMDS
-        elif start < CxoTime(cmds_recent["date"][0]) + 3 * u.day:
+        elif start < CxoTime(cmds_recent.meta["loads_start"]) + 3 * u.day:
             # Query starts near beginning of recent commands and *might* need some
             # archive commands. The margin is set at 3 days to ensure that OBS
             # command continuity is maintained (there is at least one maneuver).
@@ -418,6 +418,7 @@ def update_archive_and_get_cmds_recent(
     cmds_recent.deduplicate_orbit_cmds()
     cmds_recent.remove_not_run_cmds()
     cmds_recent = add_obs_cmds(cmds_recent, pars_dict, rev_pars_dict)
+    cmds_recent.meta["loads_start"] = start.date
 
     if cache:
         # Cache recent commands so future requests for the same scenario are fast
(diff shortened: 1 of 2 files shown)


kadi/commands/commands_v2.py
```diff
@@ -418,6 +418,7 @@ def update_archive_and_get_cmds_recent(
     cmds_recent.deduplicate_orbit_cmds()
     cmds_recent.remove_not_run_cmds()
     cmds_recent = add_obs_cmds(cmds_recent, pars_dict, rev_pars_dict)
+    cmds_recent.meta["loads_start"] = start.date
```
jeanconn, line 421: So start is CxoTime(min(loads["cmd_start"])).  Since `loads` there can include loads that are later filtered in the loop that reads into cmds_recent, is that the right time for the logic at 228?  I'd have thought you'd want the minimum time of a from-load cmd in cmds_recent.  But hard to really dig in here.

taldcroft: Excellent question which I spent some time considering. The code in the PR should work reliably. The question being asked here is "are there any applicable archive load commands which are NOT available from the recent load commands (i.e. the last 4 weeks of weekly loads)?".  Imagine these inputs:
```
start = Apr-02-2022
Loads = APR0122 (Never run due to anomaly), APR0622 (replan), APR0822,  APR1522, APR2222
cmds_recent.meta["loads_start"] = Apr-01-2022 (first command in APR0122 approx)
```
So what we care about is the first command in the applicable loads in the time span, not the first actually run load command in the span. The reason is that the archive is not going to provide additional commands here because the APR0122 commands will also not be in the archive.

This is a bit subtle (which is never a good thing), so I put this whole response into the code. Does it make sense?

jeanconn: Thanks.  That helps.

jeanconn (approved):

jeanconn (approved):

PR #268, opened 2023-01-16
Use ruff for linting and formatting

taldcroft:
## Description

Now that ruff also supports black-style formatting we can use it exclusively for both linting and formatting. The ruff formatter is about 30 times faster than black so this is a good thing.

This PR is intended as the prototype for future ruff configurations. It applies a good set of rules without going crazy.

## Interface impacts

No code impact, but this suggests updates to the standard files in `skare3/templates`.

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

Confirmed pre-commit and workflows are working with ruff format and lint issues introduced by rebase. It failed when expected. Also confirmed that the `pyrightconfig.json` file results in notebook files being ignored by pyright in VS code.


taldcroft:
I am merging this to master to allow rebasing of the other four open kadi PR's on master. This still will have a full review and if any issues arise we will fix them in another PR.

PR #273, opened 2023-01-22
Remove a testing workaround

taldcroft:
## Description

This is a clean-up item.

## Interface impacts

None

### Unit tests

- [x] Mac

- [x] Linux

### Functional tests

No functional testing.


jeanconn:
Since the flight archive has been regenerated, am I correctly understanding that this "Fix" and its removal is a no-op?  Because there are no commands to remove it isn't removing them, right?

jeanconn:
So in conversation, the commands that were "removed" by this fix were appropriate/correct but only existed in "local" not "flight" so the test would fail until regenerating the archive.  Now the test fix can be removed and these commands can be checked instead of ignored.

jeanconn (approved):

Issue #280, opened 2023-03-03
HRC 24V states need to be added to kadi states


jskrist:
for the new HRC model to run correctly it needs the 24V power states from Kadi.  The commands for these states are:

- 224PCAON
- 224PCAOF

jeanconn:
Closed by #281

PR #281, opened 2023-03-03
HRC state updates for 24V, 15V (from SCS-134), HRC_I and HRC_S

### LINKED ISSUE #279: the 15v turn on command is now in an SCS
https://github.com/sot/kadi/blob/750b92b22a8a9e4e27b8c4a8471bdd50a0f2049e/kadi/commands/states.py#L564

In order for the HRC model to run from kadi data the 15V power state needs to be read in correctly.  The 15V turn on now happens as part of SCS-134, so I think this line needs to be updated to something like:

`command_attributes = {"tlmsid": "COACTSX", "coacts1": 134}`

jskrist:
## Description
Updated states.py:

- Account for the HRC 15V power on state changing due to the `215PCAON` within SCS 134.
- Add a new `hrc_24v` state which triggers off of `224PCAON` and `224PCAOF`
- Fix the `hrc_i` and `hrc_s` states to go `OFF` based on `2IMHVOF` and `2SPHVOF`, respectively.

Fixes #279, #280 

## Interface impacts
- Adds SCS-134 as a load command that will change the `hrc_15v` state to `ON`. Since SCS-134 has never been previously commanded this has no impact.

## Testing
- Updated test_hrc_states() to include the HRC 24v state data
- Added a test of the new SCS-134 functionality using commands from the FEB0623T test loads.

### Unit tests
- [ ] Windows (JK)
- [x] Mac (TA)
- [x] Linux (JC) 

#### Plan
Test with the updated `acis_thermal_check` https://github.com/acisops/acis_thermal_check/pull/64 as noted https://github.com/acisops/acis_thermal_check/pull/64#issuecomment-1456661999.  That was run by Dan.

JC overplotted the FOT model run outputs and the last SOT model run outputs (provided in email "HRC LR cea_check and Ska3").

DEC0522P_2ceahvpt
JAN3023U_2ceahvpt
FEB0623T_2ceahvpt

(plots from https://icxc.cfa.harvard.edu/aspect/test_review_outputs/kadi-pr281/kadi_cea_review.html)


jskrist:
I don't know how to mock the kadi commands in order to insert the new SCS commanding to test that branch of the code explicitly, but would be happy to add the test if someone can help me figure out how to.

taldcroft:
@jskrist - thanks for getting started on this. There shouldn't be a need for the inner code changes you made to add in a transition due to SCS-134. You can see how this is handled in the current version with some new commits.

FYI I used this notebook as a helper in the development process. It illustrates some tricks that help:
https://gist.github.com/taldcroft/c17bbaef6ae2c9cf759f9a7dc4472580

jskrist:
@taldcroft thanks for updating the code and getting to this so quickly. I see from your changes how to mock the commands for testing this.

How do these new changes handle historical SCS-134 calls (that weren't HRC 15v on command states?  Sorry if it's obvious from your notebook, I'm on my phone this morning and the notebook it's rendering well.

taldcroft:
BTW, the 24V HRC state is a fine thing to have, but it is not physically needed for the HRC thermal modeling. See https://github.com/acisops/acis_thermal_check/pull/64#issuecomment-1454767061.

taldcroft:
@jskrist - SCS 134 has never been activated from the load commanding:
```
cmds = get_cmds()
ok = (cmds['tlmsid'] == "COACTSX") & (cmds["coacts1"] == 134)
cmds[ok]
# Empty table
```
So we're fine. During the first 22 years (pre-B-side anomaly), even if there had been SCS-134 activations in the loads it would have been benign because it would be turning on the 15V that was already on.

jskrist:
@taldcroft great, thanks for checking.

matthewdahmer:
@taldcroft The 24v state definitely does have an effect on the modeled HRC temperature. I agree that it is very difficult to fit, but in reality it does affect temperature, and if left on inadvertently would easily push the sea temperature above the observing guideline limit.
cea_short_24v
cea_long_24v

matthewdahmer:
@taldcroft I see what you meant about the 24v not affecting the model output when fitting, and agree. The time span the 24v is on is just way too short to have a real effect. That being said if it is kept on (such as by mistake), it would affect temperature.

taldcroft:
FYI @jeanconn is going to do the review of this since I've contributed commits now.

jeanconn:
I think the current implementation does not have the interface impacts called out in the description.  Correct?

taldcroft:
@jeanconn - I've updated the description.

taldcroft:
I also just force-pushed a commit update (changed commit message) to force GitHub actions to restart. It's now green.

taldcroft:
@jeanconn - the only command in SCS-134 is the HRC 15V on.

jeanconn (approved):

PR #288, opened 2023-06-24
Compare regression test quaternions with delta-quat tolerance check

jeanconn:
## Description
Compare regression test quaternions with delta-quat tolerance check instead of == test.

This fixes the issue that the calculated quaternions on fido present insignificant numeric differences from the same quaternions calculated on kady.

### Unit tests

- [x] Linux on fido at ska3-matlab 2023.4rc6

Independent check of unit tests by @taldcroft
- [x] Mac

### Functional tests

No functional testing.

### DIFF
--- a/kadi/commands/tests/test_commands.py
+++ b/kadi/commands/tests/test_commands.py
@@ -12,7 +12,6 @@ import pytest
 from astropy.table import Table, vstack
 from Chandra.Time import secs2date
 from Quaternion import Quat
-from agasc import sphere_dist
 from cxotime import CxoTime
 from testr.test_helper import has_internet
 
@@ -316,31 +315,23 @@ def test_commands_create_archive_regress(tmpdir, version_env):
                 # Path("cmds_local.txt").write_text(out)
                 assert len(cmds_flight) == len(cmds_local)
 
-            # Roughly validate NSM quaternions and remove from later comparison
-            nsm_idxs = np.flatnonzero(cmds_flight['npnt_enab'] == False) # noqa
-            for idx in nsm_idxs:
-                flight_att = cmds_flight[idx]['params']['targ_att']
-                flight_q = Quat(q=flight_att)
-                local_q = Quat(q=cmds_local[idx]['params']['targ_att'])
-                delta_point = sphere_dist(flight_q.ra, flight_q.dec,
-                                          local_q.ra, local_q.dec)
-
-                assert delta_point < 0.1
-                assert np.isclose(flight_q.roll, local_q.roll, atol=0.1)
-
-                for cmd in (cmds_local[idx], cmds_flight[idx]):
-                    del cmd['params']['targ_att']
-
-                # Not sure if it would be OK to just remove all 'prev_att'
-                # from the comparisons but here just remove the ones that
-                # have just been checked above.
-                for i in range(idx + 1, len(cmds_flight)):
-                    if 'prev_att' in cmds_flight[i]['params']:
-                        if (cmds_flight[i]['params']['prev_att'] == flight_att):
-                            del cmds_flight[i]['params']['prev_att']
-                            del cmds_local[i]['params']['prev_att']
-                        else:
-                            break
+            # Validate quaternions using numeric comparison and then remove
+            # from the == comparison below.  This is both appropriate for these
+            # numeric values and also necessary to deal with architecture
+            # differences when run on fido vs kady for example.
+            # In reality, the numeric differences only occur on "calculated"
+            # quaternions such as the NSM quaternions.
+            for idx in range(len(cmds_flight)):
+                for att_name in ['targ_att', 'prev_att']:
+                    if att_name not in cmds_flight[idx]['params']:
+                        continue
+                    flight_q = Quat(q=cmds_flight[idx]['params'][att_name])
+                    local_q = Quat(q=cmds_local[idx]['params'][att_name])
+                    dq = flight_q.dq(local_q)
+                    for attr in ("roll0", "pitch", "yaw"):
+                        assert abs(getattr(dq, attr)) < 1e-6
+                    del cmds_flight[idx]['params'][att_name]
+                    del cmds_local[idx]['params'][att_name]
 
             # 'starcat_idx' param in OBS cmd does not match since the pickle files
             # are different, so remove it.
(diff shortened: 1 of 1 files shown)


kadi/commands/tests/test_commands.py
```diff
@@ -314,6 +316,32 @@ def test_commands_create_archive_regress(tmpdir, version_env):
...
+            nsm_idxs = np.flatnonzero(cmds_flight['npnt_enab'] == False) # noqa
+            for idx in nsm_idxs:
+                flight_att = cmds_flight[idx]['params']['targ_att']
+                flight_q = Quat(q=flight_att)
+                local_q = Quat(q=cmds_local[idx]['params']['targ_att'])
+                delta_point = sphere_dist(flight_q.ra, flight_q.dec,
+                                          local_q.ra, local_q.dec)
+
+                assert delta_point < 0.1
+                assert np.isclose(flight_q.roll, local_q.roll, atol=0.1)
```
taldcroft, line 329: I think you can do `dq = flight_q.dq(local_q)` then
```
for attr in ("roll0", "pitch", "yaw"):
    assert abs(getattr(dq, attr)) < 1e-6
```
I would expect that the arch-dependent numerical issues we are catching here should be very small. Allow for 0.1 deg seems much too large. If it is really so large then we have a different problem that should be fixed.

kadi/commands/tests/test_commands.py
```diff
@@ -314,6 +316,32 @@ def test_commands_create_archive_regress(tmpdir, version_env):
...
+
+                for cmd in (cmds_local[idx], cmds_flight[idx]):
+                    del cmd['params']['targ_att']
+
+                # Not sure if it would be OK to just remove all 'prev_att'
```
taldcroft, line 334: Why not do the same `dq` comparison for `prev_att` as for `tarq_att` in a loop?

kadi/commands/tests/test_commands.py
```diff
@@ -314,6 +316,32 @@ def test_commands_create_archive_regress(tmpdir, version_env):
                 # Path("cmds_local.txt").write_text(out)
                 assert len(cmds_flight) == len(cmds_local)
 
+            # Roughly validate NSM quaternions and remove from later comparison
```
taldcroft, line 319: Comment should reflect *why* the code is here. The *what* it is doing is also useful but that is fairly evident in the code.

kadi/commands/tests/test_commands.py
jeanconn: I thought that for your regular kadi validation you open up to 4 degrees normal sun quaternions.  I didn't look much at the why on these numeric differences for the same kinds of quaternions as much as the don't care?  So I'm not sure about what is reasonable for tolerances.  However, if the tests aren't changing that's fine.

kadi/commands/tests/test_commands.py
jeanconn: I'm not entirely sure what you mean.  The 'prev_atts' should be the same as the 'targ_att', just later, so I didn't bother to do the same check, whatever it is, on the numbers.  As to why I didn't loop through and just check them all, we can do that too. 

The differences only came up on the npnt_enab == False cmds for 'targ_att' and I couldn't pull out the next commands that referenced the same off attitudes easily.

kadi/commands/tests/test_commands.py
jeanconn: The simplest thing would just be to validate _all_ the quaternions using a Quat-based check, I just wasn't as worried about simplicity for the test code.

kadi/commands/tests/test_commands.py
taldcroft: > The 'prev_atts' should be the same as the 'targ_att'

That is the hope but it needs to be verified. Basically for a single `flight` command you can think of `prev_att` and `targ_att` as two unrelated quaternions that needs to be checked against the corresponding `local` versions. So it's just something like:
```
            nsm_idxs = np.flatnonzero(cmds_flight['npnt_enab'] == False) # noqa
            for idx in nsm_idxs:
                for att_name in ('prev_att', 'targ_att'):
                    flight_att = cmds_flight[idx]['params'][att_name]
                    flight_q = Quat(q=flight_att)
                    local_q = Quat(q=cmds_local[idx]['params'][att_name])
                    ...
```
    .

jeanconn: The issue really though is that the prev_att that is the known problem does not correspond to one that matches npnt_enab == False.  So it becomes a question about if we want to use this complicated logic to isolate to just the targ_att when npn_enab == False and the prev_att when it matches a problematic targ_att, or if we want to just open this up to do a better numeric comparison instead of == for all quaternions in the rows.

kadi/commands/tests/test_commands.py
taldcroft: > I thought that for your regular kadi validation you open up to 4 degrees normal sun quaternions.

I don't know what that means. 

Fundamentally what this test is asking is: "If I run the current kadi commands update code now over a certain time range, does that match the commands that resulted from flight processing?". The only difference should be miniscule differences attributable to floating point processor differences in complex calculations. This is a sensitive test that none of the underlying code has changed in a substantive way.

If (for example) we changed `ska_sun` to provide a very slightly different sun position, then that should generate a regression failure and we decide how to handle it.

jeanconn: OK.  For "I don't know what that means."  what I was trying to say was that the only quaternions in here that are "mismatched" on fido are the derived / calculated quaternions.  Those only come up in the context of NSM quaternions.  When you validate those quaternions in kadi validation against telemetry, you only require that they be within 4 degrees of telemetry.  So I had just been thinking for the specific case of these calculated quaternions, that they don't really need to be very accurate or precise, as they aren't *used* downstream in a way that actually requiest them to be accurate and precise.  But if you just want to verify that the numerics in the regression test output are not far from the test system that's fine.

Still the question then if we should just verify all the quaternions or have custom logic for the NSM quaternions.

jeanconn: For all but the NSM quaternions, they are just read from the backstop, so there aren't numeric issues.

taldcroft (approved):
LGTM, assuming that you ran tests on the latest commit in this PR.

jeanconn:
Yes - that was what I tested before moving this from non-WIP status and asking for review.

jeanconn:
Honestly, I hadn't run the tests on kady, so I'll leave that as an exercise for integration testing.

PR #291, opened 2023-07-24
Improve SPM eclipse enable state

### LINKED ISSUE #289: `sun_pos_mon` state needs sufficient command history
For the `sun_pos_mon` state, if you set the ‘start’ time to a time during an eclipse after the EOESTECN (battery-connect command?), that EOESTECN command isn’t included in the commands used to determine the states, and then, without that command, the logic fails to construct the automatic sun position monitor state after that eclipse exit.  

I’m basically reading this out of https://github.com/sot/kadi/blob/7fe0dd5477e804a216abf388070618d86720e204/kadi/commands/states.py#L729 . So I think we probably need a kadi fix to do something determine if the start time is in an area where this could be a problem and then read back to get more commands if necessary… or something like that (like including something about it in continuity).  I note the EOESTECN is at ‘2022:109:21:27:27.034’, so setting start to exactly that works,
```
ska3-jeanconn-fido> get_chandra_states --start 2022:109:21:27:27.034 --stop 2022:110:00:00:00.000 --state-keys=sun_pos_mon
             datestart               datestop  sun_pos_mon 
 2022:109:21:27:27.034  2022:109:22:14:07.034         DISA 
 2022:109:22:14:07.034  2022:110:00:00:00.000         ENAB 
```
and one second later does not (to get the ENAB state).
```
ska3-jeanconn-fido> get_chandra_states --start 2022:109:21:27:28.034 --stop 2022:110:00:00:00.000 --state-keys=sun_pos_mon
             datestart               datestop  sun_pos_mon 
 2022:109:21:27:28.034  2022:110:00:00:00.000         DISA 
```
See also https://chandramission.slack.com/archives/C019GLX6QEP/p1689303146817899

taldcroft:
## Description

This refactors the handling for the `sun_pos_mon` state into three separate states. This was needed in order to ensure that continuity is correct for any state query start time.

Fixes #289 

## Interface impacts

Adds new states `battery_connect` and `eclipse_enable_spm`.

### Unit tests

- [x] Mac

- [x] Linux

### Functional tests

Created (in a notebook) a new validator class:
```
class ValidateSunPosMon(ValidateStateCode):
    state_name = "sun_pos_mon"
    msids = ["aopssupm"]
    plot_attrs = PlotAttrs(title="Sun position monitor", ylabel="Sun position monitor")
    min_violation_duration = 300

    def add_exclude_intervals(self):
        super().add_exclude_intervals()
        self.exclude_ofp_intervals_except(["NRML"])

    @cached_property
    def state_vals(self):
        """Convert ENAB (commanded states) to ACT (telemetry).

        The "ENAB" is an artifact of the backstop history sun position monitor states.
        This method is otherwise equivalent to the ValidateStateCode method.
        """
        states_interp = interpolate_states(self.states, self.tlm["time"])
        state_vals = states_interp[self.state_name]
        state_vals[state_vals == "ENAB"] = "ACT"
        state_vals_raw = convert_state_code_to_raw_val(state_vals, self.state_codes)
        return state_vals_raw
```
With this the SPM transitions for a 200 day span are mostly OK:
```
>>> vp = ValidateSunPosMon("2023:200", days=200)
>>> viols = vp.get_violations()
>>> print(viols)
        start                  stop        
--------------------- ---------------------
2023:045:04:46:49.525 2023:045:04:52:17.525  # Safe mode
2023:049:03:38:31.067 2023:049:04:12:08.267  # Safe mode
2023:056:21:59:54.110 2023:056:22:06:11.310  # Real diff between expected penumbra exit and actual?
```


kadi/commands/states.py
```diff
@@ -41,6 +41,9 @@
     + ["auto_npnt", "pcad_mode", "pitch", "off_nom_roll"]
 )
 
+# State keys for SIM-related transitions.
```
jeanconn, line 44: Nit, but that comment looks wrong.

kadi/commands/states.py
```diff
@@ -711,7 +714,7 @@ class SPMEnableTransition(FixedTransition):
 
     command_attributes = {"tlmsid": "AOFUNCEN"}
     command_params = {"aopcadse": 30}
-    state_keys = ["sun_pos_mon"]
+    state_keys = SPM_STATE_KEYS
```
jeanconn, line 717: This speaks to a lack of knowledge on my part, but for a non-eclipse transition, are the battery_connect and eclipse_enable_spm keys relevant?

jeanconn (approved):
Testing looks good and code makes sense to me from a "use states and continuity in a sane way to manage these things that are tricky" perspective.

jeanconn:
Related to "With this the SPM transitions for a 200 day span are mostly OK" -- what knobs would we have to twiddle to in the command google sheet if we ended up with a mismatch or wanted to fix a "mostly ok" that wasn't based on the code being incorrect?

jeanconn:
Oh, and I guess I shouldn't have Approved with an outstanding Flake 8  issue and it still probably should have that comment fixed.

jeanconn (changes requested):
I'll reset my review to "request changes" - I had that one question about relevant states and this needs flake8 update and maybe a fix to that comment.

taldcroft:
> Related to "With this the SPM transitions for a 200 day span are mostly OK" -- what knobs would we have to twiddle to in the command google sheet if we ended up with a mismatch or wanted to fix a "mostly ok" that wasn't based on the code being incorrect?

In the `Exclude Intervals` sheet of the Chandra Command events spreadsheet, we would add interval(s) to exclude with the key `sun_pos_mon`.

kadi/commands/states.py
taldcroft: This tells the code that if you are asking for (e.g.) `battery_connect` that it should include this class in the processing that looks for potential state-changing commands since `battery_connect` and `sun_pos_mon` are coupled in some way. I'm honestly not 100% sure if this is needed but it's generally a good idea for states that work in tandem.

kadi/commands/states.py
taldcroft: Fixed

kadi/commands/states.py
jeanconn: OK, that's fine.  It seemed not necessary in passing but also seems like it would be benign if not really needed (and is tested this way as fine).

taldcroft:
@jeanconn - I think I addressed your comments.

jeanconn (approved):

PR #292, opened 2023-07-31
Add validator for sun position monitor

taldcroft:
## Description

This adds a state validator for the `sun_pos_mon` state.

It also fixes a bug where telemetry values with a trailing space character (e.g. `"ACT "`) do not match the state codes.

## Interface impacts

Adds a new plot in the state validation page.

### Unit tests

- [x] Mac

- [X] Linux

### Functional tests

From within the git repo:
```
python -m kadi.scripts.validate_states
```
This generated the following output: https://icxc.cfa.harvard.edu/aspect/test_review_outputs/kadi-pr292/

### DIFF
--- a/kadi/commands/validate.py
+++ b/kadi/commands/validate.py
@@ -678,6 +678,30 @@ class ValidateHETG(ValidateGrating):
     plot_attrs = PlotAttrs(title="HETG", ylabel="HETG")
 
 
+class ValidateSunPosMon(ValidateStateCode):
+    state_name = "sun_pos_mon"
+    msids = ["aopssupm"]
+    plot_attrs = PlotAttrs(title="Sun position monitor", ylabel="Sun position monitor")
+    min_violation_duration = 300
+
+    def add_exclude_intervals(self):
+        super().add_exclude_intervals()
+        self.exclude_ofp_intervals_except(["NRML"])
+
+    @functools.cached_property
+    def state_vals(self):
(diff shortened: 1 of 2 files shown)


kadi/commands/validate.py
```diff
@@ -678,6 +678,30 @@ class ValidateHETG(ValidateGrating):
     plot_attrs = PlotAttrs(title="HETG", ylabel="HETG")
 
 
+class ValidateSunPosMon(ValidateStateCode):
+    state_name = "sun_pos_mon"
+    msids = ["aopssupm"]
+    plot_attrs = PlotAttrs(title="Sun position monitor", ylabel="Sun position monitor")
+    min_violation_duration = 300
```
taldcroft, line 685: This 300 second value is somewhat empirical from testing. It seems there is variation between the actual eclipse exit (determined by the OBC based on batteries etc) and the predicted which is in commanding. With a 300 second value most of the cases were within the limit.

jeanconn:
This looks good to me.  I note that in review of the report page I see issues with plotting negatives in the Y axis on some of the other plots which I can open as a separate issue.

taldcroft:
Thanks. If you are good with this then please approve it.

jeanconn:
Still running the unit tests.

jeanconn (approved):

jeanconn:
I think this is ready-to-go.

PR #295, opened 2023-08-15
Use cached property in commands validation

taldcroft:
## Description

This just moves to code to use a nicer pattern for cached properties. This required a slight change in when the exclude intervals are fetched from Google sheets.

The PR also makes one small out-of-scope change to remove the unused `update_tlm()` base method. That originally served a purpose but is no longer used.

## Testing

Used local ska3-performance installation.

### Unit tests

- [x] Mac (importantly the newly-available validation regression tests)
```
(ska3-perf) ➜  kadi git:(use-cached-property) git rev-parse HEAD             
87bf8716ac03e51bef60a2f53f21d609df3e67f9
(ska3-perf) ➜  kadi git:(use-cached-property) pytest kadi
==================================================== test session starts ====================================================
collected 218 items                                                                                                         

kadi/commands/tests/test_commands.py ...........................................................................      [ 34%]
kadi/commands/tests/test_states.py ......................x.............................................x............. [ 72%]
..........                                                                                                            [ 76%]
kadi/commands/tests/test_validate.py ...................                                                              [ 85%]
kadi/tests/test_events.py ..........                                                                                  [ 89%]
kadi/tests/test_occweb.py ......................                                                                      [100%]

======================================== 216 passed, 2 xfailed in 135.94s (0:02:15) =========================================
```

- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/kadi/commands/validate.py
+++ b/kadi/commands/validate.py
@@ -133,30 +133,30 @@ class Validate(ABC):
         self.start: CxoTime = self.stop - days * u.day
         self.no_exclude = no_exclude
 
+        # Get the exclude intervals from the google sheet along with any auto-generated
+        # ones. This creates self.exclude_intervals which is a Table. By virtue of `tlm`
+        # and `states` properties that get used, this also creates self.tlm and
+        # self.states.
+        self.add_exclude_intervals()
+
     def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
         if cls.state_name is not None:
             cls.subclasses.append(cls)
 
-    @property
+    @functools.cached_property
     def tlm(self):
-        if not hasattr(self, "_tlm"):
-            logger.info(
-                f"Fetching telemetry for {self.msids} between {self.start.date} and"
+        logger.info(
+            f"Fetching telemetry for {self.msids} between {self.start.date} and"
+            f" {self.stop.date}"
+        )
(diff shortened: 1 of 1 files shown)


kadi/commands/validate.py
```diff
@@ -133,30 +133,30 @@ def __init__(self, stop=None, days: float = 14, no_exclude: bool = False):
...
+    @functools.cached_property
     def tlm(self):
-        if not hasattr(self, "_tlm"):
-            logger.info(
-                f"Fetching telemetry for {self.msids} between {self.start.date} and"
+        logger.info(
+            f"Fetching telemetry for {self.msids} between {self.start.date} and"
+            f" {self.stop.date}"
+        )
+        _tlm = get_telem_table(self.msids, self.start, self.stop)
```
jeanconn, line 153: If we're moving away from using attributes directly to store these (via`CLS._attribute_names`) it feels a little odd to bother with the underscore names, as we tend not to use them generally.

taldcroft: This is to not overload the method name itself. I.e. inside `def tlm(self)`, `tlm` is defined as the unbound method. A type checker or IDE will notice stuff like this.

jeanconn: Right, I wasn't suggesting `tlm` as the replacement but fine to leave these with the current naming hinting at the property-ness of the internals.

PR #302, opened 2023-11-23
WIP: Validate finish line

taldcroft:
## Description

WIP for the following:
- [ ] Use higher precision Sun position for off_nom_roll state to remove those violations
- [ ] Fix SPM violations (either with tolerance or fixing a root-cause issue)
- [ ] Email notification to `aca` for violations
- [ ] General code tidy as warranted

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.


kadi/scripts/validate_states.py
```diff
@@ -4,13 +4,17 @@
...
 from cheta import fetch
+from cxotime import CxoTime, CxoTimeLike
 
 import kadi
-from kadi.commands import conf, validate
+import kadi.commands
+from kadi.commands.validate import Validate
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("kadi")
```
taldcroft, line 17: This makes the logger use the same output format as the rest.

kadi/scripts/validate_states.py
```diff
@@ -68,15 +205,22 @@ def main(args=None):
     maude.conf.cache_msid_queries = True
     fetch.CACHE = True
 
-    if opt.in_work:
-        conf.include_in_work_command_events = True
+    kadi.commands.conf.include_in_work_command_events = opt.in_work
```
taldcroft, line 208: In the unlikely event that a user has set the configuration in a file, the code now overrides that. I.e. the command line option wins.

kadi/scripts/validate_states.py
```diff
@@ -56,6 +67,132 @@ def get_opt():
...
+    """Send an email alert for validation violations.
+
+    Parameters
+    ----------
+    text : str
+        Text for email
+    opt : argparse.Namespace
+        Command-line options
+    """
+    from acdc.common import send_mail
```
jeanconn, line 189: acdc is presently a non-fsds package.  One of the reasons we let it go there was the idea that this acdc.common send mail import was *also* only used by other non-fsds packages.  So I see this new import in kadi validation as suggesting one of several paths:
1) Follow through and copy/move this send_mail into something like ska_helpers
2) Put acdc back into ska3-flight
3) Put this kadi validation and trending into a separate non-fsds package instead of kadi proper
4) Leave it like this and leave acdc in non-fsds packages but realize the validation reporting then depends on a package outside of flight (practically, the validation is likely going to be run from HEAD ska3/flight which will have acdc installed).

taldcroft: Good point. Let's go with 4 right now. Tests should still pass without acdc installed. I have been bothered by having this utility living in acdc so this will get fixed in a reasonable time scale.

jeanconn (approved):

kadi/scripts/validate_states.py
```diff
@@ -47,6 +51,13 @@ def get_opt():
         action="store_true",
         help="Include in-work events in validation (for checking new events)",
     )
+    parser.add_argument(
+        "--email",
+        action="append",
+        dest="emails",
+        default=[],
+        help='Email address for notification (multiple allowed, use "TEST" for testing)',
```
jeanconn, line 59: Oh, I don't know if it is useful to document in here that this TEST thing is a feature of acdc.common.send_mail where it just logs the output.

taldcroft: It will be documented when that function gets a new home in ska_helpers.

jeanconn:
Though I should have asked before I approved -- why does this need ska_helpers 0.13.0?

taldcroft:
> Though I should have asked before I approved -- why does this need ska_helpers 0.13.0?

I think it was the coerce_types thingy.

jeanconn:
Makes sense but is in 0.12.0 (0.13.0 just has a test update so I was trying to figure out if if you were calling out a different package or different release).

PR #304, opened 2023-12-01
Update code and tests for use with ska_sun accurate position

### LINKED ISSUE #270: Update kadi off_nominal_roll validation to use accurate Sun position
The off-nominal roll computation gets highly inaccurate at far tail sun when the Sun position is off by a ~0.3 deg, which is the case for the "fast". Use https://github.com/sot/ska_sun/pull/25 to use the more accurate DE432 method here.

taldcroft:
## Description

This PR takes advantage of the associated PR's below (higher precision Sun position by default) to achieve improved accuracy of the off_nom_roll state. This remove those violations from the kadi validation page and allows a tighter check tolerance.

Despite 7 files being changed, there is only one substantive update in `states.py`, and even this one is not changing the behavior at all just doing the pitch/off_nom_roll calculation in a more efficient way. 

The rest of the changes are just code tidying or adding a new development script for evaluating state speed.

Closes #270

## Required dependent PRs

These are required to get the performance/accuracy gains and pass tests.

- https://github.com/sot/ska_sun/pull/36
- https://github.com/sot/chandra_maneuver/pull/27
- https://github.com/sot/chandra_aca/pull/162

### Unit tests

- [x] Mac
```
(ska3-perf) ➜  kadi git:(improve-off-nom-roll-state-accuracy) git rev-parse HEAD
e201f4461472dd0f77cc5b5f22ee9ed06a530bb5
(ska3-perf) ➜  kadi git:(improve-off-nom-roll-state-accuracy) pytest kadi
==================================================== test session starts =====================================================
collected 219 items                                                                                                          

kadi/commands/tests/test_commands.py ...........................................................................       [ 34%]
kadi/commands/tests/test_states.py ......................x.............................................x.............. [ 72%]
.........                                                                                                              [ 76%]
kadi/commands/tests/test_validate.py ....................                                                              [ 85%]
kadi/tests/test_events.py ..........                                                                                   [ 89%]
kadi/tests/test_occweb.py ......................                                                                       [100%]

========================================= 217 passed, 2 xfailed in 94.37s (0:01:34) ==========================================
```
- [x] Linux

#### Branch
```
(ska3-perf) ➜  validate git:(improve-off-nom-roll-state-accuracy) python performance_states.py                              
kadi.__version__='7.7.2.dev15+ge201f44'
get_states took 3.4 sec
2nd get_states took 3.1 sec
```
#### Flight
```
(ska3) ➜  validate git:(improve-off-nom-roll-state-accuracy) python performance_states.py
kadi.__version__='7.6.0'
get_states took 6.2 sec
2nd get_states took 5.9 sec
```


jeanconn:
I don't see where this PR does "Improve off-nominal roll state accuracy".

kadi/commands/states.py
```diff
@@ -1102,9 +1102,13 @@ def update_sun_vector_state(cls, date, transitions, state, idx):
             current index into transitions
         """
         if state["pcad_mode"] == "NPNT":
-            q_att = Quat([state[qc] for qc in QUAT_COMPS])
-            state["pitch"] = ska_sun.pitch(q_att.ra, q_att.dec, date)
-            state["off_nom_roll"] = ska_sun.off_nominal_roll(q_att, date)
+            ra, dec, roll = state["ra"], state["dec"], state["roll"]
+            time = date2secs(date)
+            sun_ra, sun_dec = ska_sun.position(time)
```
jeanconn, line 1107: I think maybe this should just be explicit about the position method it uses instead o using the ska_sun default.

taldcroft: This is an intentional and important part of the design to delegate the behavior to `ska_sun.conf.sun_position_default_method`. That way you can choose to compute states with the fast position method.

taldcroft:
> I don't see where this PR does "Improve off-nominal roll state accuracy".

Fair enough. I have updated the title and also the description:

This PR takes advantage of the associated PR's below (higher precision Sun position by default) to achieve improved accuracy of the off_nom_roll state. This remove those violations from the kadi validation page and allows a tighter check tolerance.

kadi/commands/states.py
jeanconn: Fair.  I wasn't clear from our previous discussion (where accurate was getting faster) if kadi would still need to use both position methods in kadi.commands.states .  Should it always be accurate when we're doing validation?  I don't see that configured if so.

taldcroft: Validation should run using the defaults just like a user or application normally would. That's also an important part of the design, that validation matches the way the code gets used.

If for some reason there was a code change that updated the ska_sun default, then all kinds of tests would break so that won't happen accidentally.

jeanconn:
Do you have a conftest.py in kadi/commands/tests/ in your local repo that isn't commited to version control @taldcroft ?

taldcroft:
> Do you have a conftest.py in kadi/commands/tests

Yes, fixed.

jeanconn (approved):
My only note is that it would be great if github would display something that said "this notebook isn't going to show you any figures and you'll have to open it in something else" for the pr304-funct-test.ipynb.

taldcroft:
Hum, that's interesting about the notebook. These are plotly figures, so apparently that doesn't get saved in the notebook or doesn't render on GitHub. Good to remember.

PR #305, opened 2023-12-02
Change limits in SPM and off-nom-roll validators to fix violations

taldcroft:
## Description

This PR opens up a couple of validation limits to remove spurious violations.

For the sun position monitor, the original tolerance of 300 seconds was a little too tight and a few cases of up to 360 seconds occurred. This is related to the difference between predicted eclipse exit times and what the OBC actually did on orbit. This increases the limit to 400 sec.

For off-nominal roll, the error due to sampling the maneuver attitude at only 5-minute bins is a bit larger than was originally assumed. Increasing the tolerance during NMAN from 10 deg to 12 deg removes violations from a 200 day sample.

## Interface impacts

None.

### Unit tests

- [x] Mac (done at the first commit before fixing the notebook name)
This requires the file `kadi/commands/test/conftest.py` from [e99d43e](https://github.com/sot/kadi/commit/e99d43e8ff1371cd95ebff24e98cddbd524b2d37). This file was used in testing and inadvertently left out of the commits.
```
(ska3-perf) ➜  kadi git:(fix-spm-off-nom-roll-violations) git rev-parse HEAD      
467afb24b99214851778deb97789d5b508032034
(ska3-perf) ➜  kadi git:(fix-spm-off-nom-roll-violations) pytest
==================================================== test session starts =====================================================
collected 219 items                                                                                                          

kadi/commands/tests/test_commands.py ...........................................................................       [ 34%]
kadi/commands/tests/test_states.py ......................x.............................................x.............. [ 72%]
.........                                                                                                              [ 76%]
kadi/commands/tests/test_validate.py ....................                                                              [ 85%]
kadi/tests/test_events.py ..........                                                                                   [ 89%]
kadi/tests/test_occweb.py ......................                                                                       [100%]

========================================= 217 passed, 2 xfailed in 94.04s (0:01:34) ==========================================
```

Independent check of unit tests by Jean with that conftest.py and with a ska3-masters environment and dev/local versions of chandra_aca, ska_sun, and chandra_maneuver in PYTHONPATH. 
- [x] Linux

### Functional tests

See the included notebook.


validate/pr305-func-test.ipynb
```diff
@@ -0,0 +1,2731 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Functional testing for PR-302\n",
```
jeanconn, line 7: Should probably be PR-305 here but not sure if it is worth the update.

jeanconn:
Technically I think this needs a rebase to get that conftext.py and pass tests.

taldcroft:
> Technically I think this needs a rebase to get that conftest.py and pass tests.

`conftest.py` is in master so when this is merged the `master` branch will pass. I've updated the documentation to reflect that the testing results depend on having that file from `master` in the correct place.

jeanconn:
That's fine.  My process is always just to check out the PR and run tests; so for this I'll copy the conftest and run tests if that's the only change that is missing for tests to pass.

PR #309, opened 2023-12-25
Add "HRC not run" event to handle HRC being disabled

taldcroft:
## Description

This PR provides a better way to handle HRC being disabled than adding a bunch of `Command not run` events. It follows the pattern from `Load not run` and `Observing not run`.

I have added this new event type to the FOT MP documentation page just to get it done with now. There is no danger of this accidentally getting into the flight Events Sheet.

## Interface impacts

Supports new HRC not run entry on Events Sheet.

### Unit tests

- [x] Mac (new unit test)
```
➜  kadi git:(hrc-disable) git rev-parse HEAD                                                        
8158e773d3a87499373378d17a5e414fd669b46d
➜  kadi git:(hrc-disable) pytest
===================================================== test session starts ======================================================
collected 221 items                                                                                                            

kadi/commands/tests/test_commands.py .............................................................................       [ 34%]
kadi/commands/tests/test_states.py ......................x.............................................x................ [ 73%]
.......                                                                                                                  [ 76%]
kadi/commands/tests/test_validate.py ....................                                                                [ 85%]
kadi/tests/test_events.py ..........                                                                                     [ 90%]
kadi/tests/test_occweb.py ......................                                                                         [100%]

========================================== 219 passed, 2 xfailed in 141.11s (0:02:21) ==========================================
```

- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/kadi/commands/commands_v2.py
+++ b/kadi/commands/commands_v2.py
@@ -350,12 +350,12 @@ def update_archive_and_get_cmds_recent(
     # E.g. an SCS-107 near end of loads where next week vehicle loads only were
     # uplinked.
     not_run_loads = {}
-    for not_run_type in ("Load", "Observing"):
-        not_run_loads[not_run_type] = set(
-            cmd_event["Params"]
+    for not_run_type in ("Load", "Observing", "HRC"):
+        not_run_loads[not_run_type] = {
+            cmd_event["Params"]: cmd_event["Date"]
             for cmd_event in cmd_events
             if cmd_event["Event"] == f"{not_run_type} not run"
-        )
+        }
 
     # Update loads table and download/archive backstop files from OCCweb
     loads = update_loads(scenario, cmd_events=cmd_events, lookback=lookback, stop=stop)
@@ -375,6 +375,15 @@ def update_archive_and_get_cmds_recent(
         elif load_name in not_run_loads["Observing"]:
             # Cut observing commands
             bad = np.isin(cmds["scs"], [131, 132, 133])
+            cmds = cmds[~bad]
+        elif load_name in not_run_loads["HRC"]:
+            # Cut HRC state-changing commands
+            date = not_run_loads["HRC"][load_name]
+            bad = (
+                ((cmds["tlmsid"] == "COACTSX") & (cmds["coacts1"] == 134))
+                | ((cmds["tlmsid"] == "COENASX") & (cmds["coenas1"] == 89))
+                | ((cmds["tlmsid"] == "COENASX") & (cmds["coenas1"] == 90))
+            ) & (cmds["date"] > date)
             cmds = cmds[~bad]
 
         if len(cmds) > 0:
--- a/kadi/commands/commands_v2.py
+++ b/kadi/commands/commands_v2.py
@@ -380,7 +380,8 @@ def update_archive_and_get_cmds_recent(
             # Cut HRC state-changing commands
             date = not_run_loads["HRC"][load_name]
             bad = (
-                ((cmds["tlmsid"] == "COACTSX") & (cmds["coacts1"] == 134))
+                (cmds["tlmsid"] == "224PCAON")
+                | ((cmds["tlmsid"] == "COACTSX") & (cmds["coacts1"] == 134))
                 | ((cmds["tlmsid"] == "COENASX") & (cmds["coenas1"] == 89))
                 | ((cmds["tlmsid"] == "COENASX") & (cmds["coenas1"] == 90))
             ) & (cmds["date"] > date)
--- a/kadi/commands/tests/test_commands.py
+++ b/kadi/commands/tests/test_commands.py
@@ -1332,45 +1332,88 @@ def test_get_rltt_scheduled_stop_time():
 
 
 # For HRC not run testing
-stop_date_2023_12_20 = stop_date_fixture_factory("2023-12-20")
+stop_date_2023200 = stop_date_fixture_factory("2023:200")
 
 
 @pytest.mark.skipif(not HAS_INTERNET, reason="No internet connection")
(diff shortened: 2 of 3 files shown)


jeanconn:
As a reviewer on this I'm working just from the kadi states understanding of how HRC is commanded.  So from that perspective --  is there a name for the nominal operation of HRC via these SCS activations as compared to the hardware commands?  This ends up just being a documentation question really.

taldcroft:
> is there a name for the nominal operation of HRC via these SCS activations

I'm not aware of a good and descriptive name, but they fall within the "HRC return to science operations concept" (along with a hundred other things).

jeanconn:
OK.  Sounds fine as long as it will be clear to the users here that the HRC not run is only going to apply to this nominal operations concept and not to any hardware commands.

kadi/commands/tests/test_commands.py
```diff
@@ -1329,3 +1329,48 @@ def test_get_rltt_scheduled_stop_time():
...
+def test_hrc_not_run_scenario(stop_date_2023_12_20):  # noqa: ARG001
+    """Test custom scenario with HRC not run"""
+    from kadi.commands.states import get_states
+
+    # First make the cmd_events.csv file for the scenario
+    scenario = "hrc_not_run"
+    cmds_dir = Path(commands_v2.conf.commands_dir) / scenario
+    cmds_dir.mkdir(exist_ok=True, parents=True)
+    # Note variation in format of date, since this comes from humans.
+    cmd_evts_text = """\
```
jeanconn, line 1348: This test would probably be stronger if the states were evaluated with and without the hrc not run -- but as a sanity and regress test this is good.

kadi/commands/commands_v2.py
```diff
@@ -376,6 +376,15 @@ def update_archive_and_get_cmds_recent(
...
             bad = np.isin(cmds["scs"], [131, 132, 133])
             cmds = cmds[~bad]
+        elif load_name in not_run_loads["HRC"]:
+            # Cut HRC state-changing commands
+            date = not_run_loads["HRC"][load_name]
+            bad = (
+                ((cmds["tlmsid"] == "COACTSX") & (cmds["coacts1"] == 134))
+                | ((cmds["tlmsid"] == "COENASX") & (cmds["coenas1"] == 89))
+                | ((cmds["tlmsid"] == "COENASX") & (cmds["coenas1"] == 90))
+            ) & (cmds["date"] > date)
```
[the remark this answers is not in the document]
jeanconn: Comment for whole block of filtering could also probably use an update.

taldcroft:
>  Sounds fine as long as it will be clear to the users here that the HRC not run is only going to apply to this nominal operations concept and not to any hardware commands.

In fact the hardware commands do actually run on board, they just don't do anything because the instrument is off.

But yes, there is a bit of duct tape in all this. F_HRC_SAFING disables SCSs 92, 93 and 134, but the kadi state values depend on 89, 90, and 134 due to the way HRC is run. So this PR ignores 89, 90 instead of 92, 93, because that is what makes things work.

kadi/commands/commands_v2.py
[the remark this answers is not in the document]
taldcroft: The docs say "HRC observations following the event Date were disabled", which I would take as `>`. But in practice it does not matter, the script run timing will be accurate to maybe a minute.

kadi/commands/tests/test_commands.py
taldcroft: Good idea, I'll fix this.

taldcroft:
Thinking more, `F_HRC_SAFING` can be run in the middle of an HRC observation and it turns off HRC immediately. So I need to insert some commands at the event date to do that.

jeanconn:
"In fact the hardware commands do actually run on board, they just don't do anything because the instrument is off."

Right, I meant hardware commands in the loads, as those are the ones that are defined as state-changing-things in kadi.

jeanconn:
"So I need to insert some commands at the event date to do that."

Sounds good to take the time to do that update to this PR to future-proof it.

taldcroft:
> I meant hardware commands in the loads, as those are the ones that are defined as state-changing-things in kadi.

OK now I get your drift. There are dozens of HRC hardware commands in the loads. But you were referring to just the two state-changing ones that control the 15V and 24V. Those are never commanded in the loads in the Return to science concept and (if they were included) there would be no way to stop them from running besides running SCS-107. So I think were good on this.

taldcroft:
@jeanconn - thanks for the comments, this PR is definitely better now and hopefully ready!

kadi/commands/commands_v2.py
```diff
@@ -376,6 +376,16 @@ def update_archive_and_get_cmds_recent(
             # Cut observing commands
             bad = np.isin(cmds["scs"], [131, 132, 133])
             cmds = cmds[~bad]
+        elif load_name in not_run_loads["HRC"]:
+            # Cut HRC state-changing commands
+            date = not_run_loads["HRC"][load_name]
+            bad = (
+                (cmds["tlmsid"] == "224PCAON")
```
jeanconn, line 383: OK so I'm a little confused by this.  Should the command be removed here or should Hrc24vOn_Transition be updated so it is a no-op if the 15v is off?  Or does it not matter?  Not trying to overcomplicate here but not sure what will make the most sense to us later.  Thanks!

taldcroft: Updating `Hrc24vOn_Transition` like you said might work and might be slightly more faithful to what is really happening. That said, doing a state-dependent transition is complicated and will require spending several more hours on this code that might never even get used. So I'm going to say that perfect is the enemy of good here, I've already invested too much time on this.

jeanconn (approved):

PR #311, opened 2024-01-11
Skip tests if internet is not available

javierggt:
## Description

Skip tests if internet is not available.

## Interface impacts

None

### Unit tests

- [x] Mac (with WiFi on and off by @taldcroft)
```
masters) ➜  kadi git:(skip-test) git rev-parse HEAD
b47f543e7f61acd180fda7b649e0358c8411f39b
```
### Without internet (wifi off)
```
(masters) ➜  kadi git:(skip-test) pytest
===================================================== test session starts =====================================================
collected 174 items                                                                                                           

kadi/commands/tests/test_commands.py ........sssss..ss...............................................ss...s......       [ 43%]
kadi/commands/tests/test_states.py ......................x.......................                                       [ 70%]
kadi/commands/tests/test_validate.py .sssssssssssssssssss                                                               [ 81%]
kadi/tests/test_events.py ..........                                                                                    [ 87%]
kadi/tests/test_occweb.py ssssssssssssssssssssss                                                                        [100%]

========================================= 122 passed, 51 skipped, 1 xfailed in 16.44s =========================================
```
### With internet
```
(masters) ➜  kadi git:(skip-test) pytest
===================================================== test session starts =====================================================
collected 228 items                                                                                                           

kadi/commands/tests/test_commands.py .................................................................................. [ 35%]
..                                                                                                                      [ 36%]
kadi/commands/tests/test_states.py ......................x.............................................x............... [ 73%]
........                                                                                                                [ 77%]
kadi/commands/tests/test_validate.py ....................                                                               [ 85%]
kadi/tests/test_events.py ..........                                                                                    [ 90%]
kadi/tests/test_occweb.py ......................                                                                        [100%]

========================================= 226 passed, 2 xfailed in 100.15s (0:01:40) ==========================================
```

### Functional tests

No functional testing.

### DIFF
--- a/kadi/commands/tests/test_validate.py
+++ b/kadi/commands/tests/test_validate.py
@@ -10,13 +10,19 @@ import pytest
 import ska_sun
 
 from kadi.commands.utils import compress_time_series
-from kadi.commands.validate import Validate, ValidateRoll
+from kadi.commands.validate import Validate, ValidateRoll, get_command_sheet_exclude_intervals
 
 # Regression testing for this 5-day period covering a safe mode with plenty of things
 # happening. There are a number of violations in this period and a couple of excluded
 # intervals.
 REGRESSION_STOP = "2022:297"
 REGRESSION_DAYS = 5
+
+try:
+    get_command_sheet_exclude_intervals()
+    CMD_SHEET_AVAILABLE = True
+except ConnectionError:
+    CMD_SHEET_AVAILABLE = False
 
 
 def write_regression_data(stop, days, no_exclude):
@@ -107,6 +113,7 @@ def test_validate_subclasses():
     assert set(data_all_exp.keys()) == {cls.state_name for cls in Validate.subclasses}
 
 
+@pytest.mark.skipif(not CMD_SHEET_AVAILABLE, reason="Command sheet not available")
 @pytest.mark.parametrize("cls", Validate.subclasses)
 @pytest.mark.parametrize("no_exclude", [False, True])
 def test_validate_regression(cls, no_exclude, fast_sun_position_method):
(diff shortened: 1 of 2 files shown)


taldcroft (approved):

taldcroft:
@javierggt - I tested this by turning off my wifi and found some other failures. In addition, the validation tests should succeed if there is internet. If the command sheet disappeared for some reason then we definitely want tests to fail.

javierggt:
I suppose turning off wifi introduces extra failures. I checked that this PR skipped the failing tests, so tests pass on GRETA.

If we want to have tests fail if the command sheet is not reachable, then I suppose the only solution is to explicitly skip on GRETA, but I am assuming that having a list of GRETA machines is not what you want. Do we have a test that would tell us we are on GRETA?

taldcroft:
> If we want to have tests fail if the command sheet is not reachable, then I suppose the only solution is to explicitly skip on GRETA, but I am assuming that having a list of GRETA machines is not what you want. Do we have a test that would tell us we are on GRETA?

`has_internet()` checks that `https://google.com` is reachable. So I think if that is true then all the command states functionality is expected to work. All the commands V2 tests (which depend on the same Google Sheet) use `has_internet()` as the skipif test, so applying this to validation makes sense to me. If the machine can reach google.com but not the Command Sheet, then we have a big problem.

The only risk in this strategy would be if HEAD was failing `has_internet()` since that would mean silently skipping tests that should really fail. But this is not a credible situation at least for any extended time.

We don't have a test that says we are on GRETA, the goal in unit tests is to ask whether required resources are available for a particular test. We do have a test for being on HEAD which gets used in a few places where this helps (https://github.com/search?q=org%3Asot%20on_head_network&type=code).

PR #312, opened 2024-01-17
Fix unintended side-effect of add_cmds

taldcroft:
## Description

The `CommandTable.add_cmds()` method had a bug where it would update the original `self` table in place (using `remove_rows`) prior to using it for computing the new table with additional commands. I think the original idea of this method was for the whole thing to be in-place but that was inconvenient.

This fixes the issue by not using `remove_rows` and defines a new reference to the original commands which is filtered NOT in-place.

## Interface impacts

None.

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/kadi/commands/core.py
+++ b/kadi/commands/core.py
@@ -608,16 +608,17 @@ class CommandTable(Table):
         -------
         CommandTable
         """
+        orig = self
         if rltt is not None:
-            remove_idxs = np.where(self["date"] > rltt)[0]
-            self.remove_rows(remove_idxs)
+            ok = orig["date"] <= rltt
+            orig = orig[ok]
 
         try:
             # Substantially faster than plain Table vstack (this is slow due to
             # checks for strings overflowing and other generalities)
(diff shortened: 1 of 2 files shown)


javierggt (approved):
I reviewed the changes and they make sense to me. I read the test in detail, and verified that the test passes in the branch and fails in master.

While reading the test I noticed that the commands that are clipped are the ones with times after the RLTT, which was not obvious to me before. The docstring does not specify that, and instead uses colloquial language (_Clip existing commands to the RLTT_). It would have been nice to have it in the docstring).

taldcroft:
> The docstring does not specify that, and instead uses colloquial language (Clip existing commands to the RLTT). It would have been nice to have it in the docstring).

I'll improve the docstring, that's a good point. There is already some PR cross-talk with the docstring here so I don't want to touch this now, praying to not generate merge conflicts when this gets merged.

PR #319, opened 2024-02-09
Fix bug in OBS obs_stop for cmd_evt maneuver and improve default_stop handling

taldcroft:
### OBS obs_stop

This fixes a bug where an `OBS` statement coming from command events (during anomaly recovery) could have the wrong `obs_stop` value. 

When an `OBS` statement is created it uses the time of the next maneuver or schedule stop to set `obs_stop`. In normal processing those are always available, but for a maneuver in anomaly recovery neither are available and it uses a default 7 days in the future.

Normally commands which change in any way by subsequent events or loads are detecting by the diff algorithm and replace. However the `obs_stop` update is just part of the command `params` and this is not checked in the diffing. Doing so introduces false diffs due to small floating point precision issues. So this PR special-cases `OBS` so that the `params` are included in the match key.

### KADI_COMMANDS_DEFAULT_STOP

An unrelated issue surfaced during testing that the `KADI_COMMANDS_DEFAULT_STOP` environment variable was not being applied as a filter to events the command events sheet. This was unexpected to me and would make it very difficult to do detailed "scenario testing" of stepping through the timeline of past events. In this case having the full history of command events meant that the code always had subsequent maneuvers so seeing the OBS problem was difficult.

## Interface impacts

None

### Unit tests

- [x] Mac
```
(ska3) ➜  kadi git:(fix-matching-for-obs-cmds) git rev-parse HEAD  
3301432d93e8dbd79418873e076abab68a1a9170
(ska3) ➜  kadi git:(fix-matching-for-obs-cmds) pytest
=========================================================== test session starts ===========================================================
collected 232 items                                                                                                                       

kadi/commands/tests/test_commands.py ......................................................................................         [ 37%]
kadi/commands/tests/test_states.py .......................x..............................................x.......................   [ 77%]
kadi/commands/tests/test_validate.py ....................                                                                           [ 86%]
kadi/tests/test_events.py ..........                                                                                                [ 90%]
kadi/tests/test_occweb.py ......................                                                                                    [100%]

============================================================ warnings summary =============================================================
kadi/kadi/commands/tests/test_commands.py::test_get_starcats_obsid
  /Users/aldcroft/miniconda3/envs/ska3/lib/python3.10/site-packages/mica/utils.py:16: FutureWarning: ska_load_dir() is deprecated. Use parse_cm.paths.load_dir_from_load_name()
    dir_parts = ska_load_dir(load_name).parts

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
========================================== 230 passed, 2 xfailed, 1 warning in 99.43s (0:01:39) ===========================================
```
- [x] OSX

```
(ska3) flame:kadi jean$ pytest
===================================================================== test session starts =====================================================================
rootdir: /Users/jean/git, configfile: pytest.ini
collected 232 items                                                                                                                                           

kadi/commands/tests/test_commands.py .......s.......s.....................................................................s                             [ 37%]
kadi/commands/tests/test_states.py .......................x..............................................x.......................                       [ 77%]
kadi/commands/tests/test_validate.py ....................                                                                                               [ 86%]
kadi/tests/test_events.py ..........                                                                                                                    [ 90%]
kadi/tests/test_occweb.py ......................                                                                                                        [100%]

====================================================================== warnings summary =======================================================================
kadi/kadi/commands/tests/test_commands.py::test_get_starcats_each_year[year0]
  /Users/jean/miniconda3/envs/ska3/lib/python3.10/site-packages/setuptools_scm/git.py:295: UserWarning: git archive did not support describe output
    warnings.warn("git archive did not support describe output")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=============================================== 227 passed, 3 skipped, 2 xfailed, 1 warning in 81.41s (0:01:21) ===============================================
(ska3) flame:kadi jean$ git rev-parse HEAD
3301432d93e8dbd79418873e076abab68a1a9170
```

### Functional tests

Detailed functional testing in `validate/pr319`. This demonstrates that the patch fixes the problem. This functional validation should work on HEAD.

```
# I did this before committing the script. To replicate copy the script to a different name.
git switch master  
python run_updates.py > master.log
git switch fix-matching-for-obs-cmds
python run_updates.py > fix-matching-for-obs-cmds.log
```


kadi/commands/commands_v2.py
```diff
@@ -1315,26 +1315,12 @@ def get_matching_block_idx(cmds_arch, cmds_recent):
...
-    arch_vals = [
-        tuple(
-            row[x].decode("ascii") if isinstance(row[x], bytes) else str(row[x])
-            for x in key_names
-        )
-        for row in cmds_recent
-    ]
-
-    diff = difflib.SequenceMatcher(a=recent_vals, b=arch_vals, autojunk=False)
+    diff = difflib.SequenceMatcher(a=arch_vals, b=recent_vals, autojunk=False)
```
taldcroft, line 1323: Note that the original code had the names reversed (`recent` vs `arch`). `arch_vals` was coming from `cmds_recent` and `recent_vals` was from `cmds_arch_recent`. The names are fixed now with no change in content of `a` and `b`.

jeanconn: Great!  I was trying to get why the right/newer answer wasn't "winning" even if there was a bug that meant the code wasn't recognizing the obs statement was different.  I think this fix explains that.

kadi/commands/commands_v2.py
```diff
@@ -1362,6 +1367,32 @@ def get_matching_block_idx(cmds_arch, cmds_recent):
     return idx0_arch, idx0_recent
 
 
+def get_list_for_matching(cmds: CommandTable) -> list[tuple]:
```
jeanconn, line 1370: Seems like there should be a new unit test for this function / bugfix.

kadi/commands/tests/test_commands.py
```diff
@@ -372,7 +372,8 @@ def stop_date_fixture(monkeypatch):
     return stop_date_fixture
 
 
-stop_date_2021_10_24 = stop_date_fixture_factory("2021-10-24")
+# 2021:297 0300z just after recovery maneuver following 2021:296 NSM
+stop_date_2021_10_24 = stop_date_fixture_factory("2021-10-24 03:00:00")
```
jeanconn, line 376: Though I'm guessing this is constructed to hit the OBS bug?

kadi/commands/commands_v2.py
taldcroft: This is covered in the functional testing. The issue with adapting that functional test into a unit test is that it requires a snapshot of how the commands archive looked on 2024-Feb-04. This is not possible to construct from the current flight archive because the Not-run commands have all be removed.

In theory the test could check for the presence of this 100 Mb file in $SKA/data/kadi and run if available. Let me know if you are not satisfied with the one-time functional test. My thinking was that given the code documentation it seems not credible that future updates will break this particular fix.

kadi/commands/tests/test_commands.py
taldcroft: This is unrelated to the OBS bug.

The original test was passing only because of the issue where all command events (including those after the stop date) were included. This change includes additional event(s) that are required for the original test to pass.

kadi/commands/commands_v2.py
jeanconn: OK, when I wrote my comment I was thinking it would not be that bad to generate short cmds archive test data to show that OBS statement is correctly updated by new processing. But if you think that doesn't have much value that's fine.

jeanconn (approved):

jeanconn:
The only thing that niggles at me is that KADI_COMMANDS_DEFAULT_STOP is not really a default (if it has a value, we're doing something special that is non-default) but that is neither here nor there.

PR #322, opened 2024-02-16
WIP : Change time threshold for auto-enable of SPM following eclipse

taldcroft:
## Description

The matching time threshold of 125 seconds between battery connect time and eclipse entry appears to be too low. Eclipses on 2024:043 and 2024:046 had a delta of 129 to 130 seconds. This results in the `sun_pos_mon` remaining disabled coming out of eclipse.

This PR changes that value to 150 sec (2.5 minutes). **This value may be changed since it causes test fails.**

### Unit tests

- [ ] Mac

- [ ] [PLATFORM]:

### Functional tests

For about 30 seconds I changed the status of the two `COMMAND_SW | TLMSID=AOFUNCEN, AOPCADSE=30` commands on day 43 and 46 to be `In-work`. With that in place I ran kadi validate states locally and confirmed that the `sun_pos_mon` state matches telemetry.

### DIFF
--- a/kadi/commands/states.py
+++ b/kadi/commands/states.py
@@ -782,7 +782,7 @@ class SPMEclipseEnableTransition(BaseTransition):
     Automatic enable of sun position monitor.
 
     This occurs 11 minutes after eclipse exit, but only if the battery-connect
-    command occurs within 2:05 minutes of eclipse entry.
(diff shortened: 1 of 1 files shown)


taldcroft:
This is not the problem.

PR #361, opened 2025-06-13
Try shorter regress file names for windows

jeanconn:
## Description

Fixes #

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.


jeanconn:
Is this closer to what you had in mind Tom?  I didn't actually see a cheat sheet for remaking the regression data that is in here so I just renamed the files while also making a code change that seems like it would support the convention.

jeanconn:
And it looks like we just got ruff 0.12.0.  "Yay".

taldcroft:
Yeah, we should globally disable `PLC0415 `import` should be at the top-level of a file`. There are often good reasons for doing this.

